PR #13, opened 2020-10-25
Fixing lisp_fn macro functionality to work with new pdumper logic.

DavidDeSimone:
This PR repairs the ability to use the #[lisp_fn] macro to define Rust functions callable from emacs. Previously in remacs, we called "xmalloc", and left the pointer linger on the heap for the duration of the program. Changes to the emacs build process (including pdump and unexec) have made this no longer viable. Instead I have opt'd to use the MaybeUninit construct to make an actual aligned and uninitialized struct that we can fill with data and hand off to the C layer.  

Basic functionality test can be performed via the following patch applied to this branch:

```
diff --git a/rust_src/src/example.rs b/rust_src/src/example.rs
new file mode 100644
index 0000000000..d91ad58bf2
--- /dev/null
+++ b/rust_src/src/example.rs
@@ -0,0 +1,9 @@
+use lazy_static::lazy_static;
+use remacs_macros::lisp_fn;
+
+#[lisp_fn]
+pub fn eptn(t: crate::lisp::LispObject) -> crate::lisp::LispObject {
+    crate::lisp::LispObject::from(true)
+}
+
+include!(concat!(env!("OUT_DIR"), "/example_exports.rs"));
diff --git a/rust_src/src/lib.rs b/rust_src/src/lib.rs
index bd9efb24b5..3016f87696 100755
--- a/rust_src/src/lib.rs
+++ b/rust_src/src/lib.rs
@@ -31,6 +31,7 @@ extern crate remacs_macros;
 mod remacs_sys;
 #[macro_use]
 mod lisp;
+mod example;

 #[cfg(not(test))]
 include!(concat!(env!("OUT_DIR"), "/c_exports.rs"));
```

### DIFF
--- a/rust_src/remacs-macros/lib.rs
+++ b/rust_src/remacs-macros/lib.rs
@@ -1,4 +1,4 @@
-#![recursion_limit = "128"]
+#![recursion_limit = "256"]
 
 extern crate lazy_static;
 extern crate proc_macro;
@@ -72,6 +72,7 @@ pub fn lisp_fn(attr_ts: TokenStream, fn_ts: TokenStream) -> TokenStream {
     let cname = lisp_fn_args.c_name;
     let sname = concat_idents("S", &cname);
     let fname = concat_idents("F", &cname);
+    let srname = concat_idents("SR", &cname);
     let rname = function.name;
     let min_args = lisp_fn_args.min;
     let mut windows_header = quote! {};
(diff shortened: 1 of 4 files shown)


DavidDeSimone:
Callout for this PR - you may notice a variable called "windows_header" that is still constructed based off the size of Lisp_Subr and not Aligned_Lisp_Subr. I have not tested this patch on windows and have not investigated the requirements of "windows_header" now that we are using Aligned_Lisp_Subr.

PR #15, opened 2020-10-26
Removing need for 'use lazy_static' in a module that uses lisp_fn

DavidDeSimone:
Currently to use lisp_fn, you need to include `use lazy_static::lazy_static;` in your module. This PR removes this requirement via a custom use alias per invocation of lisp_fn.


brotzeit:
LGTM

PR #17, opened 2020-11-03
[WIP] Pipe between a rust thread and the emacs process

DavidDeSimone:
Opening a draft PR to share my development with the community and open it up for comments.

Idea is that this is not what the final interface will look like, but more that I wanted to share my POC for exchanging data between a rust thread and the lisp thread.

I have yet to profile performance, that is coming soon. 

To test run the following lisp

``` elisp
(setq pipe-proc (make-pipe-process :name "piped-buffer"))
(thread-path pipe-proc)
(process-send-string pipe-proc "Hello")
```

Specifying a :filter option to make-pipe-process allows us to define a function that receives the input from the input of the pipe. The default filter is to just dump the input into the buffer associated with the pipe-process. 

That :filter function will be our primary way to get that data into the lisp world for usage.


rust_src/src/json_rpc.rs
```diff
@@ -0,0 +1,63 @@
...
+    WRITE_TO_SUBPROCESS = 1,
+    READ_FROM_SUBPROCESS = 2,
+    SUBPROCESS_STDOUT = 3,
+    READ_FROM_EXEC_MONITOR = 4,
+    EXEC_MONITOR_OUTPUT = 5,
+}
+
+#[lisp_fn]
+pub fn thread_path(proc: LispObject) -> bool {
+    let proc_raw = unsafe { XPROCESS(proc) };
```
brotzeit, line 25: I'm trying to find the code that allows `LispProcessRef` as argument to avoid this line.

DavidDeSimone:
Update:

I've added a macro to generate interaction with rust async functions. Now a rust async function can be marked with the "async_stream" macro

```
pub async fn my_stream(data: String) -> String {
    // ... perform my async calculations
}
```

In lisp, we can do something like the following:

```
(defun handler (process results) (print results)) ; or do whatever we want back in lisp land
(setq pipe (call-my-stream 'handler)) ; The result of call-my-stream is an emacs pipe process that will invoke your handler
(async-send-message pipe "my data....")
```

Currently the design is limited to strings. The next step is to allow for transfer of user data offthread, with a return call when calculation is complete.

PR #18, opened 2020-11-06
restore vector functions

brotzeit:


rust_src/src/lisp.rs
```diff
@@ -106,11 +142,33 @@ impl<T> ExternalPtr<T> {
...
+    type Target = T;
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*self.0 }
+    }
 }
 
-pub type LispSubrRef = ExternalPtr<Aligned_Lisp_Subr>;
-impl LispSubrRef {
-    pub fn as_mut(self) -> *mut Aligned_Lisp_Subr {
-	self.0
```
brotzeit, line 114: @DavidDeSimone I added it to `ExternalPtr<T>`

rust_src/src/lisp.rs
```diff
@@ -106,11 +142,38 @@ impl<T> ExternalPtr<T> {
...
+impl<T> DerefMut for ExternalPtr<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        unsafe { &mut *self.0 }
     }
 }
+
+// Lisp_Subr support
+
+pub type LispSubrRef = ExternalPtr<Aligned_Lisp_Subr>;
+unsafe impl Sync for Aligned_Lisp_Subr {}
```
brotzeit, line 179: This was in the old code base ?

rust_src/src/lisp.rs
DavidDeSimone: That looks good

rust_src/src/lisp.rs
DavidDeSimone: There was an impl Sync for Lisp_Subr in the old codebase, however that lived in remacs_sys.rs. I did add equivalent code for Aligned_Lisp_Subr implementing Sync in remacs_sys, so I don't think we need that here. We should still have `pub type LispSubrRef = ExternalPtr<Aligned_Lisp_Subr>;` here however

rust_src/Cargo.toml
```diff
@@ -58,9 +58,7 @@ panic = "abort"
 
 [features]
 default = [
-  "use-xml2", "unexec", "window-system", "window-system-x11", "libvterm",
-
-]
+  "use-xml2", "unexec", "window-system", "window-system-x11",]
```
DavidDeSimone, line 61: Do we want to remove libvterm here intentionally?

brotzeit: Yeah, it's currently not in the code base.

rust_src/src/lisp.rs
brotzeit: > I did add equivalent code for Aligned_Lisp_Subr implementing Sync in remacs_sys, so I don't think we need that here

Ah I missed that.

PR #31, opened 2020-12-15
Add a `webrender` window system

harryfei:
* Add new option `--with-webrender` in `configuration.ac` 
* Add new cargo feature `windows-system-webrender` in rust_src

It's buildable while the result binary isn't runnable.

### DIFF
--- a/rust_src/src/wrterm.rs
+++ b/rust_src/src/wrterm.rs
@@ -1,0 +1,172 @@
+//! wrterm.rs
+
+use std::ptr;
+
+use remacs_macros::lisp_fn;
+
+use crate::{
+    frame::LispFrameRef,
+    lisp::{ExternalPtr, LispObject},
+    remacs_sys::{
+        font, image, wr_display_info, wr_output, Display, Emacs_Pixmap, Pixmap, Qnil, WRImage,
+        Window, XrmDatabase,
+    },
+};
+
+#[repr(transparent)]
+pub struct DisplayInfoRef(*mut wr_display_info);
+unsafe impl Sync for DisplayInfoRef {}
+
+pub type OutputRef = ExternalPtr<wr_output>;
+pub type DisplayRef = ExternalPtr<Display>;
+pub type ImageRef = ExternalPtr<WRImage>;
+
+#[no_mangle]
+pub static tip_frame: LispObject = Qnil;
+
+#[no_mangle]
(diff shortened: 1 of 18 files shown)


brotzeit:
Can you please add instructions to the readme.

rust_src/src/wrterm.rs
```diff
@@ -0,0 +1,182 @@
...
+
+#[allow(unused_variables)]
+#[no_mangle]
+pub extern "C" fn wr_get_baseline_offset(output: OutputRef) -> i32 {
+    unimplemented!();
+}
+
+#[allow(unused_variables)]
+#[no_mangle]
+pub extern "C" fn wr_get_pixel(ximg: ImageRef, x: i32, y: i32) -> i32 {
```
DavidDeSimone, line 68: For these extern "C" functions, is it intentional to type the arguments as i32 vs a type from libc (e.g. libc::c_int)?

src/wrterm.h
```diff
@@ -0,0 +1,160 @@
...
+struct wr_display_info
+{
+  /* Chain of all wr_display_info structures.  */
+  struct wr_display_info *next;
+
+  /* The generic display parameters corresponding to this NS display.  */
+  struct terminal *terminal;
+
+  /* This is a cons cell of the form (NAME . FONT-LIST-CACHE).  */
+  Lisp_Object name_list_element;
```
DavidDeSimone, line 24: Are there any concerns for GC tracing of `name_list_element` here? Or do we expect the currently existing codebase to handle this?

harryfei:
> Can you please add instructions to the readme.

Currently, this branch isn't runnable. To avoid confusing newcomers with it's state, I'd like to add build instructions after it becomes runnable. 😄

rust_src/src/wrterm.rs
harryfei: Why I use "i32" instead of "c_int" here is to avoid `as` casts in function body. IMHO, `libc::c_int` equals `i32`.

src/wrterm.h
harryfei: The existing codebase needs this field in `xxx_display_info` struct. I think they can handle it properly.

brotzeit (approved):

rust_src/src/wrterm.rs
brotzeit: I remember that using `i32` caused issues in certain situations, but I don't remember why.

PR #34, opened 2020-12-26
remove unexec features for rust

brotzeit:
I think the portable dumper should be stable enough so we can drop the unexec modifications we made for remacs. But maybe I'm wrong.


DavidDeSimone (approved):
LGTM

PR #35, opened 2020-12-28
webrender: implement x-create-frame

harryfei:


rust_src/src/wrterm.rs
```diff
@@ -133,12 +145,78 @@ pub extern "C" fn x_focus_frame(frame: LispFrameRef, noactivate: bool) {
     unimplemented!();
 }
 
+// FRAME is used only to get a handle on the X display.  We don't pass the
+// display info directly because we're called from frame.c, which doesn't
+// know about that structure.
+#[no_mangle]
+pub extern "C" fn x_get_focus_frame(frame: LispFrameRef) -> LispObject {
```
brotzeit, line 152: I think you have to use `*mut frame` here.

harryfei: We defined those in rust code.
```rust
#[repr(transparent)]
pub struct ExternalPtr<T>(*mut T);

pub type LispFrameRef = ExternalPtr<Lisp_Frame>; 

```

So `LispFrameRef` just equals `*mut frame`.

brotzeit: I noticed that we used the C types in remacs so I thought we did it for a reason.

harryfei: I think it's safe. Some functions in remacs use this trick too.

https://github.com/remacs/remacs/blob/master/rust_src/src/dispnew.rs#L65
```rust
/// Redraw frame FRAME.
#[no_mangle]
pub extern "C" fn redraw_frame(mut frame: LispFrameRef) {
```

harryfei:
Merge? @brotzeit

brotzeit (approved):

PR #37, opened 2020-12-29
Additional JS Features

DavidDeSimone:
Features:
* Javascript related README for general documentation
* Javascript tests - can be run via `./src/emacs --batch -l ./test/js/bootstrap.el`
* Adds let, with-temp-buffer, with-current-buffer
* Updated json-ser logic to throw error when bad data is passed in instead of unwrapping. 
* Fixing bug with return value of lambdas
* Fixing bug where lisp.list() was being evaluated in lisp.setq calls. 
* Adding basic WebWorker support - allows for parallel execution of javascript. NOTE: Lisp functions cannot be called from WebWorkers, however Deno can still be used.  
* Improved JS lambda cleanup logic
* Improved JS lambda creation performance


JS_README.md
```diff
@@ -0,0 +1,115 @@
...
+    console.log('hello ' + a);
+    lisp.print(b);
+    return 3;
+}, 3, lisp.make.alist({x: 3}));
+```
+
+The user can also define functions via `defun` that will call back into javascript. Like in normal defun, DOCSTRING and the INTERACTIVE decl are optional. This example features both docstring and interactive
+
+``` js
+lisp.defun("my-function", "This is my cool function", {interactive: true, args: "P\nbbuffer"}, (buff) => console.log("Hello Buffer"));
```
yyoncho, line 69: this is too ugly. Can we have something like this:

```
lisp.defun.my_function(a) { 
     }.interactive(....)
      .comment("foo-bar")
```
(or similar builder pattern).

JS_README.md
```diff
@@ -0,0 +1,115 @@
...
+
+``` js
+let buffer = lisp.get_buffer_create("mybuf");
+```
+
+The lisp object uses reflection to invoke the equivalent of `(get-buffer-create "mybuf")`. In this example, get-buffer-create returns a buffer object. In order to represent this, we use a technique called proxying, in which we give a reference to the elisp object to an internal field of a javascript object. This field is only accessible by the native layer, so it cannot be tampered with in the javascript layer. When javascript calls subsequent functions using `buffer`, a translation layer will extract the elisp object and invoke it on the function.
+
+Primative data structures will be converted automatically. More complex data structures can be converted to native javascript objects via a call to a special function called `json`
```
yyoncho, line 11: I was thinking about that - seems like for the performance purpose we might need 2 versions of the functions - one that automatically converts and one that uses the original values to avoid redundant conversions. It is not something to address know, but if will be good to know that we can achieve that in the future.

yyoncho (approved):
I haven't tested the code, just reviewed the docs. IMHO better to have JS_README.md as part of readme.md for better visibility.

JS_README.md
DavidDeSimone: One JS pattern I like is named argumemnts. What do you think of:

``` js
lisp.defun({
    name: "my-function",
    docstring: "My Docstring",
    interactive: true, 
    func: () => { },
});
```

yyoncho: Looks good.

JS_README.md
DavidDeSimone: We can definitely do that, but I agree I think its something to slate for a future release.

JS_README.md
DavidDeSimone: Added.

JS_README.md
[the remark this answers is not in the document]
DavidDeSimone: Fixed

DavidDeSimone:
I've moved JS_README to README.md, under a Javascript subsection.

PR #42, opened 2021-01-01
Adding Async File I/O for module loading / Adding eval-js-buffer and eval-js-region

DavidDeSimone:
Additions per title.

### DIFF
--- a/rust_src/src/javascript.rs
+++ b/rust_src/src/javascript.rs
@@ -59,6 +59,11 @@ impl EmacsJsRuntime {
     }
 }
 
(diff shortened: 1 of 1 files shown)


yyoncho (commented):
Looks good. I was expecting eval-js-* function to be implemented based on `eval-js(string)`.

yyoncho (approved):

DavidDeSimone:
> Looks good. I was expecting eval-js-* function to be implemented based on `eval-js(string)`.

Good point, we can get better code reuse that way. I will address and merge.

PR #60, opened 2021-01-05
Update README.md

DavidDeSimone:
As we build up momentum and features, I think it's good to update the readme with information based on how to build, and why the user should use emacs-ng. I've partially based this on remacs README.md with some tweaks. We we add new features like git2-rs, and stabilize features like webrender, we will update the README showing the power behind those features too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,11 @@ The goal of this fork is to explore new development approaches.
 
 # Why Emacsng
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,24 +1,158 @@
...
+- [emacs-ng](#emacs-ng)
+    - [Motivation](#motivation)
+- [Why Emacsng](#why-emacsng)
+    - [Why JavaScript](#why-javascript)
+        - [Performance](#performance)
+- [Getting Started](#getting-started)
+    - [Requirements](#requirements)
+    - [Building emacsng](#building-emacsng)
+    - [Running emacsng](#running-emacsng)
+    - [Features](#features)
```
brotzeit, line 15: Should be one level up right ?

DavidDeSimone: Updated

brotzeit:
I had to download the submodules, can't find it in the readme though.

DavidDeSimone:
> I had to download the submodules, can't find it in the readme though.

Good Catch! Updated.

PR #46, opened 2021-01-01
webrender: render basic text

harryfei:

PR #62, opened 2021-01-06
add lisp crate

brotzeit:
It's not ready, but I will hurry to get this done. Maybe we have to make some methods to functions because you can only add impls for types that are defined in the same crate.

I think we shouldn't merge anything before this is done.

Afterwards we have to adjust the build system since it seems generation of lisp_fn only works under the current path. And we certainly want to define lisp functions in different crates.


brotzeit:
I kept the module name remacs_sys for now because it's easier to make the changes. Changing this afterwards is easier.

DavidDeSimone (approved):
Changed 'crate' to '$crate' in eval_macros and everything just worked. Patch passes JavaScript tests and elisp tests - LGTM.

brotzeit:
Thanks, it was too late for me yesterday I guess =)

But I still have to activate the webrender code. I will do this in my lunch break.

brotzeit:
We still need to update the imports for the webrender modules. However I would like to do it together with the rebasing of #46.

Issue #63, opened 2021-01-06
Remove uses of `static mut`


ubolonton:
`static mut` is extremely hard to use correctly: https://github.com/rust-lang/rust/issues/53639.

We should use [once_cell](https://github.com/matklad/once_cell) instead. (It can also replace our uses of `lazy_static!`.)

DavidDeSimone:
https://github.com/emacs-ng/emacs-ng/pull/64 addressed our JavaScript layer's usage of static mut, however there are still a number of uses in the codebase, including the code generated from `#[lisp_fn]`, the `Lisp_Fwd` struct, and almost all of bindgen. bindgen will be harder to tackle, but we can still work the other cases.

PR #61, opened 2021-01-06
Adding Docstrings, fixing color formatting issues.

DavidDeSimone:
In addition, I am defining some special forms used for a js-repl project.

PR #64, opened 2021-01-06
Remove Static Mut usage in javascript.rs 

DavidDeSimone:
Per https://github.com/emacs-ng/emacs-ng/issues/63, we should avoid using mut static due to it's difficulty of being used correctly. 

I have opt'd instead to use `thread_local!`, creating a struct bundling the mut statics together in a safe to access RefCell. 

I branched this off of https://github.com/emacs-ng/emacs-ng/pull/61, so you will see some unrelated changes until that lands. The changes here are all to javascript.rs, along with some better testing coverage for the case of calling a function with the flow JS -> Lisp -> JS -> Lisp -> JS


ubolonton:
`thread_local!` means each thread will have its own `EmacsMainJsRuntime`. Do we want that?

DavidDeSimone:
I think that we do, this being my though process - though if I am missing something please just let me know: 

Lisp has the ability to spawn concurrent threads, which share global variables. While this gives the user intuition that JS globals will be shared, I am anticipating issues with how Deno/v8 will respond to that situation, even if I guard the shared state with a Mutex. The `MainWorker` references a v8 isolate, which has the restriction that `An isolate can be entered by at most one thread at any given time. The Locker/Unlocker API must be used to synchronize.` [0]

If we run the following code, it will produce an error:

```lisp 
(eval-js "console.log('foo')")
(make-thread (lambda () (eval-js "console.log('bar')")))
```

However, I can get this code to work by modifying this patch slightly to remove the `std::once::ONCE` I used for tokio completion. Users still can set lisp variables if they want to share state between lisp threads. The only thing I think would be nice is to inherent the settings of (js-initialize) in that case, but that should be a trivial lisp function to define. 

[0] https://v8docs.nodesource.com/node-0.8/d5/dda/classv8_1_1_isolate.html

ubolonton:
> Lisp has the ability to spawn concurrent threads, which share global variables. While this gives the user intuition that JS globals will be shared, I am anticipating issues with how Deno/v8 will respond to that situation, even if I guard the shared state with a Mutex. The `MainWorker` references a v8 isolate, which has the restriction that `An isolate can be entered by at most one thread at any given time. The Locker/Unlocker API must be used to synchronize.` [0]

Does this means each Deno isolate is subjected to a "global interpreter lock", like ELisp/Python/Ruby?

> 
> If we run the following code, it will produce an error:
> 
> ```lisp
> (eval-js "console.log('foo')")
> (make-thread (lambda () (eval-js "console.log('bar')")))
> ```
> However, I can get this code to work by modifying this patch slightly to remove the `std::once::ONCE` I used for tokio completion. 

Do you mean we get an error without this PR (1 global js runtime), and no error with it (1 js runtime per-thread)? I didn't get an error either way. What's the error you got? 

> Users still can set lisp variables if they want to share state between lisp threads. The only thing I think would be nice is to inherent the settings of (js-initialize) in that case, but that should be a trivial lisp function to define.
> 

IIUC, the trade-off is not having global js state (only thread-local), so that we don't have to make Deno's global lock and ELisp's global lock work together?

DavidDeSimone:
> Does this means each Deno isolate is subjected to a "global interpreter lock", like ELisp/Python/Ruby?

My understanding is its not really 'global' lock, it's more thread specific. An isolate is designed for use on a thread, and should not be shared between threads without being very careful about it. 

> Do you mean we get an error without this PR (1 global js runtime), and no error with it (1 js runtime per-thread)? I didn't get an error either way. What's the error you got?

I get an error without this PR, and do not get an error with this PR. On my system, without this PR, I see

```
foo
"Uncaught RangeError: Maximum call stack size exceeded"
```

> IIUC, the trade-off is not having global js state (only thread-local), so that we don't have to make Deno's global lock and ELisp's global lock work together?

Yes, that is the tradeoff. I like it because the code is simpler, and  I believe will be less prone to error. The end user can still share data between lisp threads in JS, it will just need to do so via lisp variables.

ubolonton:
> My understanding is its not really 'global' lock, it's more thread specific. An isolate is designed for use on a thread, and should not be shared between threads without being very careful about it.

Yeah, "global" may sound misleading. In the case of Deno, it's scoped to the isolate, but it appears to be the same rule: evaluating code requires acquiring an exclusive lock, so at any single time there is only one thread (per isolate) running JS.  

> I do need to add a `sleep-for 20` after that call to allow for the lisp thread to run, running this with
> `emacs --batch -l javascript.el`
> 

That's interesting. I'll try to reproduce it with a sleep.

> Yes, that is the tradeoff. I like it because the code is simpler, and I believe will be less prone to error. The end user can still share data between lisp threads in JS, it will just need to do so via lisp variables.
> 
> EDIT: And for clarity, I want to note that this is completely unrelated to WebWorkers and JavaScript threads. This PR only effects the interaction of Deno with Lisp spawned threads.

I think this is a big downside. Another potential issue is each isolate has its own heap and GC. That sounds a bit too heavy.

I think there are at least 2 alternatives:
- Use a single dedicated thread for the JS runtime, which never executes ELisp code.
- Ensure that each thread acquires/releases the JS lock anytime it acquires/releases the Elisp lock.

However, I'm not familiar enough with v8's working to map out more details, so I can't tell whether these can work. I think we can live with thread-local JS runtimes for now.

PR #79, opened 2021-01-10
unify file name style

harryfei:
* unify file name style
* change file permission of some files


DavidDeSimone (approved):
Works on my machine, LGTM

PR #80, opened 2021-01-10
rename remacs_sys module

brotzeit:


harryfei:
Can we have a new name instead of `generated`?

brotzeit:
Sure what do you have in mind ?

yyoncho:
IMHO `remacs_sys` should be `lisp` or `emacs`.

harryfei:
I like `sys` or `emacs`

brotzeit:
Keep in mind that the generated stuff will be pulled into another crate.

brotzeit:
And that the generated bindings also contain non-lisp stuff.

brotzeit:
I'm working on a PR that also covers this.

PR #85, opened 2021-01-13
webrender: render colored and styled text

harryfei:


rust_src/crates/lisp/src/lisp.rs
```diff
@@ -99,8 +99,8 @@ impl From<usize> for LispObject {
 }
 
 impl From<i32> for LispObject {
-    fn from(o: i32) -> Self {
-        LispObject::from_C(o as EmacsInt)
+    fn from(v: i32) -> Self {
+        Self::from_fixnum(EmacsInt::from(v))
```
DavidDeSimone, line 103: This is a good change. I'll test that it doesn't effect JS, but it shouldn't

rust_src/src/webrender_backend/term.rs
```diff
@@ -235,6 +332,37 @@ extern "C" fn frame_visible_invisible(frame: *mut Lisp_Frame, is_visible: bool)
     }
 }
 
+extern "C" fn set_background_color(f: *mut Lisp_Frame, arg: LispObject, _old_val: LispObject) {
```
brotzeit, line 335: Can we use `mut frame: LispFrameRef` ?

harryfei: Unfortunately, we can't.
This function is referenced in this below code. So the type checker need it's signature to be current one.

```rust
    let handlers: [frame_parm_handler; 47] = [
        None,
        None,
        None,
        None,
        None,
        Some(set_background_color),
        None,
        None,
        None,
        None,
        None,
        None,
```

rust_src/crates/lisp/src/lisp.rs
DavidDeSimone: Follow up: It doesn't effect JS, good to go.

PR #88, opened 2021-01-14
configure native comp and webrender in ci

brotzeit:


harryfei:
We can have 2 job in CI.  One is for building `emacs-ng` without webrender and another is for building with `webrender`.

brotzeit:
Do we really need this ? Why not test everything together ?

harryfei:
If we build with `webrender` option, `gtk` or `macosx` gui related code will not be built. We need to continue to check those graphic backends' code.

brotzeit:
Hmm, yeah I forgot you are also modifying code that's related to this.

brotzeit:
I opened another pull request that replaces this one (#200). After we have merged it, we can also include webrender.

PR #89, opened 2021-01-14
provide git-init lisp function by using git2rs as dependency

brotzeit:


DavidDeSimone (approved):
LGTM, this is exciting!

Issue #93, opened 2021-01-15
fails to install rustup toolchain


petr-tik:
Hey, awesome project - v excited about having first-class async support in emacs. 

I cloned and tried to build on my linux (xubuntu 18.04) desktop, where I've previously compiled native-comp from source. However, I got a failure when trying to install the rust-toolchain specified in the file. 

Possibly a rustup bug rather than emacs-ng. Thought it might be worth reporting. 

```bash
petr_tik@merluzon:/tmp/emacs-ng master $
rustup toolchain install nightly-2020-09-20-x86_64-unknown-linux-gnu
warning: downloading with complete profile isn't recommended unless you are a developer of the rust language
info: syncing channel updates for 'nightly-2020-09-20-x86_64-unknown-linux-gnu'
info: latest update on 2020-09-20, rust version 1.48.0-nightly (f68e08933 2020-09-19)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'llvm-tools-preview'
info: downloading component 'miri'
info: downloading component 'rls'
info: downloading component 'rust-analysis'
info: downloading component 'rust-analyzer-preview'
info: downloading component 'rust-docs'
info: downloading component 'rust-src'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustc-dev'
info: downloading component 'rustc-docs'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: using up to 500.0 MiB of RAM to unpack components
info: installing component 'clippy'
info: installing component 'llvm-tools-preview'
 20.0 MiB /  20.0 MiB (100 %)  16.6 MiB/s in  1s ETA:  0s
info: installing component 'miri'
info: installing component 'rls'
info: installing component 'rust-analysis'
info: installing component 'rust-analyzer-preview'
info: installing component 'rust-docs'
 13.2 MiB /  13.2 MiB (100 %)  13.1 MiB/s in  1s ETA:  0s
info: installing component 'rust-src'
info: installing component 'rust-std'
 21.4 MiB /  21.4 MiB (100 %)  14.9 MiB/s in  1s ETA:  0s
info: installing component 'rustc'
 55.3 MiB /  55.3 MiB (100 %)  16.0 MiB/s in  3s ETA:  0s
info: installing component 'rustc-dev'
 92.6 MiB /  92.6 MiB (100 %)  15.2 MiB/s in  6s ETA:  0s
info: installing component 'rustc-docs'
[...]

PR #96, opened 2021-01-16
Updating Rust nightly version

DavidDeSimone:
Per the discussion https://github.com/emacs-ng/emacs-ng/issues/93. I think we should deconflict this with https://github.com/emacs-ng/emacs-ng/pull/95 just so we don't merge them back to back and we end up a stay SEGFAULT and not know what caused it. I think merging 95 first, letting that settle, and following up with this makes sense. 

@harryfei I also touched webrender to fix some warnings about temporary values that are introduced in newer Rust.


brotzeit (approved):

brotzeit:
Ready for merge ?

PR #103, opened 2021-01-17
webrender: handle keyboard event

harryfei:
* handle keyboard event
* scroll window support
* show window cursor

### DIFF
--- a/rust_src/src/webrender_backend/cursor.rs
+++ b/rust_src/src/webrender_backend/cursor.rs
@@ -1,0 +1,120 @@
+use super::output::OutputRef;
+
+use lisp::{
+    remacs_sys::{
+        draw_glyphs_face, draw_phys_cursor_glyph, get_phys_cursor_geometry, get_phys_cursor_glyph,
+        glyph_row, glyph_type,
+    },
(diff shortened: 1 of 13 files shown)


rust_src/src/webrender_backend/cursor.rs
```diff
@@ -0,0 +1,120 @@
...
+    if unsafe {
+        (*cursor_glyph).type_() == glyph_type::XWIDGET_GLYPH
+            || (*cursor_glyph).type_() == glyph_type::IMAGE_GLYPH
+    } {
+        return;
+    }
+
+    let face = unsafe {
+        let face_id = (*cursor_glyph).face_id();
+        let face_id = std::mem::transmute::<u32, lisp::remacs_sys::face_id>(face_id);
```
DavidDeSimone, line 67: Is there anyway to avoid mem::transmute here?

brotzeit (approved):

rust_src/src/webrender_backend/cursor.rs
harryfei: No, face_id is an enum. We can only convert u32 to enum by mem::transmute.

PR #104, opened 2021-01-17
Fixing Panic when sometimes calling interactive functions

DavidDeSimone:
If you define a function in JavaScript that looks something like:

```js
lisp.defun({
    name: "foo",
    func: () => setInterval(() => console.log("Bar"), 0)
});
```

It will cause a sometimes cause a panic, because our setInterval may attempt to immediately call into the tokio runtime, despite the fact that we haven't entered the tokio runtime. There is no reason not to enter the runtime anytime we call into JS.


yyoncho (approved):
Looks good!

Issue #105, opened 2021-01-18
Tokio can only read 16 KB per file per event loop tick


DavidDeSimone:
Per https://github.com/tokio-rs/tokio/issues/1976 Tokio can only read 16KB per buffer read of file at a time. This can lead to performance degradations since we are limited by how often we tick the event loop to move promises forward. I will look into a work around.

This may also have performance implications for Deno as well.

DavidDeSimone:
I have a fix for this, however it involves changing Deno. Since we have already forked Deno, I intend on including it in our fork, however I did submit an issue upstream (https://github.com/denoland/deno/issues/9149), and I will see if they want a PR.

I've also determined that our default tick rate of 0.1s is likely too large - I am getting much better performance from a lower tickrate.

DavidDeSimone:
Fixed by https://github.com/emacs-ng/emacs-ng/pull/106

PR #106, opened 2021-01-18
Reduce unneeded polls to event loop

DavidDeSimone:
In current emacs-ng, once you initialize JavaScript, the event loop is always polling at `js-tick-rate` intervals. However, this is not needed if there are no pending module imports or promises. This just wastes CPU cycles and battery power. This patch brings us to a state of "you pay for what you use", and reduces timer usage. 

Instead, we now tick only when there are pending promises, and we increase the tick rate from 0.1 to 0.001. This addresses https://github.com/emacs-ng/emacs-ng/issues/105, and brings async performance inline with deno native, who has a much faster event loop. The user still has the ability to control the js-tick-rate if they want to lower it.

If people on lower end machines complain about 0.001 we can evaluate that default. 

Callout - Before, to cleanup lisp proxies, I ran a timer every 2.5s to sweep WeakRefs. This was kind of a hack, so I've edited the C to instead call that function when lisp's garbage collector is run.

### DIFF
--- a/rust_src/Cargo.lock
+++ b/rust_src/Cargo.lock
@@ -842,7 +842,7 @@ dependencies = [
 [[package]]
 name = "deno"
 version = "1.6.3"
-source = "git+https://github.com/DavidDeSimone/deno?branch=emacs-ng#5debde9ff0553af405fe37dc9188c47a8464370f"
+source = "git+https://github.com/DavidDeSimone/deno?branch=emacs-ng#f2aa5ffd13ca5b87c41410177ba0971c3e5469be"
 dependencies = [
  "atty",
  "base64 0.12.3",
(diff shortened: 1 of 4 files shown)


yyoncho:
Out of curiosity, can we loop in rust thread and notify the emacs thread via pipe? The advantage of this approach is that we are going to keep the UI thread off the work and that we can block in the rust thread.

DavidDeSimone:
The issue is that the event loop polls can still call lisp, for example:

```js
setInterval(() => lisp.insert('a'), 1000)
```

That lambda will be executed once every second, and will be executed from the event loop poll. If we were to poll in a separate Rust thread, that lisp call would need to be async

yyoncho:
I mean having something like this:

rust thread -> deno give me the task you want to execute
rust thread -> you have a task, let me notify emacs thread 
emacs thread -> perform the task. 

I guess that will work if you can do 1.

DavidDeSimone:
If I am understanding you currently - as the code is written right now, that would be difficult. Only because getting pending operations and executing them aren't separate operations in the current Deno API. Once I call `poll_event_loop` it will execute any ready pending promises, and then tell me if I am complete with the current event loop or if there are more promises pending.

yyoncho:
Ok, then my proposal won't work.

yyoncho (approved):

PR #113, opened 2021-01-19
remove ChangeLog files

brotzeit:
We also did this in remacs and I don't see any reason why we should keep them.


brotzeit:
I'm not sure where to document this change like you suggested @DavidDeSimone

DavidDeSimone:
Eh the more I think about it; the less I think it needs documentation. I think I was being overeager to document.

DavidDeSimone (approved):
LGTM

PR #115, opened 2021-01-20
Allow tick rate to be controlled from scripting layer directly

DavidDeSimone:
The default tick rate of 0.001 now seems to fast for me and is causing hitching. I'm going to lower it, and in an attempt to solve the issue, expose the ability to set the tick rate directly.

### DIFF
--- a/rust_src/src/javascript.rs
+++ b/rust_src/src/javascript.rs
@@ -1561,7 +1561,7 @@ pub fn js_get_tick_rate() -> LispObject {
 /// to improve file read performance, decrease to preventing
 /// hitching and save battery life.
 #[lisp_fn]
-pub fn js_set_tick_rate(f: LispObject) -> LispObject {
+pub fn js_set_tick_rate(f: LispObject) {
     let mut options = EmacsMainJsRuntime::get_options();
 
     unsafe {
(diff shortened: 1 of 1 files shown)


rust_src/src/javascript.rs
```diff
@@ -1550,6 +1550,29 @@ fn handle_error(e: std::io::Error, handler: LispObject) -> LispObject {
...
+pub fn js_get_tick_rate() -> LispObject {
+    let options = EmacsMainJsRuntime::get_options();
+    unsafe { lisp::remacs_sys::make_float(options.tick_rate) }
+}
+
+/// Sets F to be the current js tick rate. Increase this number
+/// to improve file read performance, decrease to preventing
+/// hitching and save battery life.
+#[lisp_fn]
+pub fn js_set_tick_rate(f: LispObject) -> LispObject {
```
brotzeit, line 1564: Can we use `EmacsInt` here ?

rust_src/src/javascript.rs
```diff
@@ -1550,6 +1550,29 @@ fn handle_error(e: std::io::Error, handler: LispObject) -> LispObject {
...
+    let mut options = EmacsMainJsRuntime::get_options();
+
+    unsafe {
+        lisp::remacs_sys::CHECK_NUMBER(f);
+        options.tick_rate = lisp::remacs_sys::XFLOATINT(f);
+        EmacsMainJsRuntime::set_options(options);
+    }
+
+    lisp::remacs_sys::Qnil
```
brotzeit, line 1573: I think we can drop this line if we return nothing.

DavidDeSimone: Corrected.

brotzeit (approved):

Issue #118, opened 2021-01-20
Upgrade to Deno 1.7


DavidDeSimone:
Features are outlined here:

https://deno.land/posts/v1.7

Largest change for us is that they are upgrading to tokio 1.0, which could be big for us.

yyoncho:
> which could be big for us.

Why?

DavidDeSimone:
In terms of immediate benefits, we may see some performance increases (I have not profiled yet). 

In terms of long term benefits, we will be utilizing a more stable interface for Tokio, which has some large features planned on their [2021 roadmap](https://tokio.rs/blog/2020-12-tokio-1-0), including usage of the new [io_uring interface](https://kernel.dk/io_uring.pdf).

DavidDeSimone:
Fixed by https://github.com/emacs-ng/emacs-ng/pull/161

PR #132, opened 2021-01-24
Add explicit command to run the example code

kidd:
Making the command explicit helps beginners.  The code is not valid
js, so `eval-js-buffer` would error.

### DIFF
--- a/adv-features.md
+++ b/adv-features.md
@@ -27,7 +27,7 @@ WebWorkers communicate via two special functions, postMessage and onmessage. onm
 
 Create a new file with the following code:
 
-```js
+```ts
 declare var lisp: any;
 
 const worker = new Worker(new URL("web-worker.js", import.meta.url).href,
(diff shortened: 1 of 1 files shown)


DavidDeSimone:
Good Recommendation - thanks for the patch!

DavidDeSimone (approved):

kidd:
First of all, thanks for this initiative. Best of luck!

Following the webworker example was a bit confusing for a total newbie like me.

- I've created #132 to make the type of file explicit.  
- `web-worker.js` has to be in the same path where emacs-ng was started (emacs-ng crashes if it tries to start a worker from a wrong filepath). I thought it was `default-directory`, but it's not. any hints?
- when I run the example, the full circle takes around 4 seconds to complete if the file is eval'd as ts. making it js-compatible the example is instant, so maybe I should change it to `js`?

DavidDeSimone:
* That is likely a behavior due to the JS framework that we are using. By default, it will use the directory emacs-ng was launched in. I can change that example to have it instead look in `default-dir` however, since that's something we can modify our code to do. 
* Another option is to turn off typescript type checking by running `(js-initialize :no-check t)`. That allows you to run typescipt without eating the check cost.

DavidDeSimone (approved):

DavidDeSimone:
@kidd and re; the crashing, I will need to see if that is fixed by https://github.com/emacs-ng/emacs-ng/issues/118. If you would like to make an issue for that, feel free. javascript should never cause a panic, and if it does, its a bug in my mind.

PR #147, opened 2021-01-31
Adds eval-js-expression and eval-js-literally

DavidDeSimone:
This PR adds a very requested feature, which is the equivalent of M-: but for JavaScript. This also adds `eval-js-literally`, which will actually return the value of the JavaScript expression, unlike eval-js-* which always returns nil. This will auto-translate proxys, meaning that things like:

`(eval-js-literally "lisp.current_buffer()")` works how you would hope it does - it will actually return the current buffer as a lisp object. 

It seems like we should bind `eval-js-expression` to a key binding, but I'm unsure what a good default, if any, would be. I'm open to suggestions. 

These functions will ONLY work with JavaScript. Adding TS support won't be hard, but I want to wait until https://github.com/emacs-ng/emacs-ng/issues/118 lands, because Deno 1.7 reworked their interface to compile TS at runtime (Deno.compile vs Deno.emit).

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -35,8 +35,7 @@ Now we go back to \*scratch\* and run
 
 This is a relative filepath, and it works off of emacs current working directory (cwd). If you are in doubt to what emacs cwd is, just run the lisp function `(pwd)` in the lisp scratchpad
 
-You should see "Hello TypeScript" printed. All of the eval-js* functions return nil. If you want to use a calculated value from JavaScript/TypeScript in Lisp, you can either set a variable (via `lisp.setq`, or call a lisp function with it as an argument.
-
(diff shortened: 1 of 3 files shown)


jcs090218 (approved):
Definitely a good start on eval-js! :)

yyoncho:
Are there enough use cases for the JS related functions that do not return result to justify their existence?

DavidDeSimone:
> Are there enough use cases for the JS related functions that do not return result to justify their existence?

The main difference is that `(eval-js ...)` is technically a module, so it can use import/export. Using import in eval-js-literally will throw. The other difference is that eval-js-literally is in the global context, where as a module is not. 

I think the major difference to the end user is the import/export piece. Unfortunately for us, Deno does not let me get the result of the last expression of a module, so I'm kind of stuck having two evaluation functions.

yyoncho (approved):

yyoncho:
Sounds reasonable to me. Maybe this should be added to the readme?

DavidDeSimone:
I added it to "Getting Started", but it is something low level enough we might want to add it to the README.

PR #152, opened 2021-02-02
[WIP] add more remacs code

brotzeit:
This is some code that I need for git stuff. The remacs macro `call` combined with `intern` are good testing purposes.

It seems they removed `check_cons_list` upstream. Can anybody tell if this is crucial for `funcall`.


rust_src/crates/lisp/src/eval.rs
```diff
@@ -26,3 +36,94 @@ pub fn signal_rust(error_symbol: LispObject, data: LispObject) -> ! {
...
+    InvalidFun,
+    VoidFun,
+}
+
+/// Call first argument as a function, passing remaining arguments to it.
+/// Return the value that function returns.
+/// Thus, (funcall \\='cons \\='x \\='y) returns (x . y).
+/// usage: (funcall FUNCTION &rest ARGUMENTS)
+#[allow(unused_assignments)]
+pub fn funcall(args: &mut [LispObject]) -> LispObject {
```
DavidDeSimone, line 55: Here's my take on this - I think we will get ourselves into trouble long term if we copy too much of remacs re-implementations. I think copying remacs wrapper classes, and simple one liners (like intern below) are good, but larger functions like funcall will cause issues. Instead, I think it would be better if we made these functions 'safe wrappers' around the native C functions - that way we keep the implementations aligned, but we are still writing idiomatic Rust and aren't littering the code with unsafe and `as_mut_ptr()` calls .

brotzeit: Yeah I was also hesitant to do this. I'll look for an alternative.

brotzeit:
@DavidDeSimone I removed funcall.

brotzeit:
Oh I forgot that I still need to find an alternative for funcall =)

brotzeit:
Oh I forgot that I still need to find an alternative for funcall =)

rust_src/crates/lisp/src/thread.rs
```diff
@@ -0,0 +1,25 @@
...
+    lisp::ExternalPtr,
+    remacs_sys::{current_thread as current_thread_pointer, thread_state},
+};
+
+pub type ThreadStateRef = ExternalPtr<thread_state>;
+
+pub struct ThreadState {}
+
+impl ThreadState {
+    pub fn current_buffer_unchecked() -> LispBufferRef {
```
DavidDeSimone, line 16: since this is `unchecked` do we want to mark it unsafe? What it is doing seems pretty unsafe to me.

appetrosyan: I'd say no. It doesn't help anyone. 

Usually marking a function `unsafe` communicates to the caller that they need to enforce a certain invariant or they will cause UB. You're not meant to tinker with the `current_thread_pointer` anyway, so safe rust will not cause UB in this function. 

The same goes for the other function.

appetrosyan (approved):

PR #166, opened 2021-02-11
Fixing errors not being thrown by eval-js-literally

DavidDeSimone:
Repreoduction:

```lisp
(eval-js-literally "throw new Error('err')") 
```

Above will not throw on current. In addition, this will cause a panic on current:

```js
lisp.eval_js_literally('3')
```

Both are fixed by this patch


brotzeit (approved):
LGTM

PR #169, opened 2021-02-13
[WIP] add some git interfaces

brotzeit:
I'm currently working on getting magit to work with these interfaces. I think for now it's sufficient to work with advices(I think @ubolonton mentioned this idea earlier). I hope I get a working solution for magit-blame the next days.


brotzeit:
I'll put the functions into #176

PR #171, opened 2021-02-14
javascript.rs cleanup

DavidDeSimone:
~~WIP, dependent on https://github.com/emacs-ng/emacs-ng/pull/170 landing first~~

I might add things to this list between now and that PR landing, but this is my starting off point. This PR shouldn't actually change any behavior, it should just be a refactor. 

Goals:

- [x] Remove mem::transmute usage
- [x] Rename the (horrible) naming of the get_v8_stacked_handle and get_v8_raw_handle functions 
- [x] Better documents the assumptions of the v8_handle stack that must be maintained
- [x] Remove duplication across inner_invokation and execute
  - [x] Rename execute and inner_invokation to better reflect what both are actually doing 
- [x] Reduce boilerplate of binding global functions with a macro 
- [x]  Rename init_once to init_tokio
- [x] Rename js_initialize_inner to something more generic since we call it outside of js_initialize

### DIFF
--- a/rust_src/src/javascript.rs
+++ b/rust_src/src/javascript.rs
@@ -332,6 +332,10 @@ impl EmacsMainJsRuntime {
     }
 }
 
+fn is_interactive() -> bool {
+    unsafe { !lisp::remacs_sys::globals.noninteractive1 }
+}
+
 // (DDS) This exists for breaking Deno's cache busting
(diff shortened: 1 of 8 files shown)


brotzeit (approved):
LGTM

PR #172, opened 2021-02-14
Restructure generated source to make emacsng work with rust-analyzer

yyoncho:
Seems like rust analyzer cannot handle exported symbols from `include!`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -301,3 +301,7 @@ src/gdb.ini
 /var/
 
 rust_src/Cargo.toml
+
+rust_src/crates/lisp/src/bindings.rs
+rust_src/crates/lisp/src/globals.rs
+rust_src/crates/lisp/src/definitions.rs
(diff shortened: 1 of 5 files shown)


DavidDeSimone:
It looks like our build process doesn't like this change. I think the issue is that the generated files aren't available yet when we call rustfmt. I think there are a few different ways to solve this, an easy one would be to run rustfmt after we actually build emacs. This breaks 'failing fast' in the case that someone doesn't run rustfmt though.

yyoncho:
A possible option is to commit the generated files, skip that step from the build process and regenerate the files when bumping emacs.

DavidDeSimone:
> A possible option is to commit the generated files, skip that step from the build process and regenerate the files when bumping emacs.

I think this is also a good solution.

brotzeit:
> A possible option is to commit the generated files, skip that step from the build process and regenerate the files when bumping emacs.

I already talked with @yyoncho about it some time ago and I don't see any reason why we shouldn't do it. It would also reduce our build time which is definitely an advantage.

brotzeit:
These changes were merged in #201

PR #176, opened 2021-02-21
add build option for libgit and remove libvterm

brotzeit:


configure.ac
```diff
@@ -4235,16 +4234,11 @@ fi
...
-dnl ### Use libvterm (-lvterm) if available
-HAVE_LIBVTERM=no
-if test "${with_libvterm}" != "no"; then
-  EMACS_CHECK_MODULES([LIBVTERM], [vterm])
-  if test "${HAVE_LIBVTERM}" = "yes"; then
-    AC_DEFINE(HAVE_LIBVTERM, 1, [Define to 1 if libvterm library is present.])
-  fi
+### Use libgit2 and git2rs
+if test "${with_libgit}" = "yes"; then
+  HAVE_LIBGIT=yes
```
brotzeit, line 4239: This shouldn't be necessary right ?

brotzeit:
Review, anybody ? =)

configure.ac
DavidDeSimone: You can change the conditional below to be "${with_libgit}" = "yes" if you want to avoid setting HAVE_LIBGIT=yes, but I also think what you have is fine.

DavidDeSimone (approved):
LGTM

PR #177, opened 2021-02-22
Return early from CursorMoved if top_frame is none

CarlQLange:
Running with `--with-webrender`, emacs would panic on startup with an "unwrap on a none value" message referencing the `CursorMoved` event handler. I'm not well-versed in Rust but it seems like there was an assumption in the code that `top_frame.as_frame()` would be something like `Option<LispFrameRef>` but it was in fact `None`. It looks like some of the other event handlers handled this issue in the way that I've now done so in this commit. This change has now allowed me to build a working emacs with webrender, which is quite cool.


harryfei (approved):
LGTM

brotzeit:
@CarlQLange Thanks!

CarlQLange:
Thank you!

PR #180, opened 2021-03-15
Docs: Fix features link in README.md

mcky:
Current link to `/features` 404s, based off the nav on the site it appears the path should be `/main-features`


jcs090218 (approved):
Thanks!

PR #182, opened 2021-03-17
fix required version

jackmac92:
With `8` installed I got the error 
```
Please try installing libgccjit or similar package.
If you are sure you want Emacs compiled without elisp native compiler, pass
  --without-nativecomp
to configure.
```
with `9` installed it works


jcs090218 (approved):

jcs090218:
Thanks for contributing to Emacs NG!

kiennq:
Some how `libgccjit-9-dev` is never working for me. Has to use `libgccjit-10-dev` instead.
You can refer to the script to build + make deb package for it here https://github.com/kiennq/emacs-ng/blob/master/build_emacs_ng.sh

Issue #184, opened 2021-03-17
Compile with webrender error


black7375:
**Related**
 #148 (maybe)

**Environment**
- `Linux ***-pc 5.10.18-1-MANJARO #1 SMP PREEMPT Tue Feb 23 18:07:15 UTC 2021 x86_64 GNU/Linux`

**Description**
Webrender is not set up well.

For other configurations, as shown below, it is successful.
```shell
./configure --prefix=/usr --mandir=/usr/share/man --sysconfdir=/etc --libexecdir=/usr/lib --localstatedir=/var \
            --with-sound=alsa --with-modules --with-wide-int --with-xft --with-harfbuzz --with-cairo \
            --with-x-toolkit=gtk3 --with-xwidgets \
            --without-gconf --without-gsettings \
            --with-nativecomp
```

Config Result:
```shell
Configured for 'x86_64-pc-linux-gnu'.

  Where should the build process find the source code?    .
  What compiler should emacs be built with?               gcc -g3 -O2
  Should Emacs use the GNU version of malloc?             no
    (The GNU allocators don't work with this system configuration.)
  Should Emacs use a relocating allocator for buffers?    no
  Should Emacs use mmap(2) for buffer allocation?         no
  What window system should Emacs use?                    x11
  What toolkit should Emacs use?                          GTK3
  What libraries will lib-remacs link against?            -ldl -lm -lrt
  Where do we find X Windows header files?                Standard dirs
  Where do we find X Windows libraries?                   Standard dirs
  Does Emacs use -lXaw3d?                                 no
  Does Emacs use -lXpm?                                   yes
  Does Emacs use -ljpeg?                                  yes
  Does Emacs use -ltiff?                                  yes
  Does Emacs use a gif library?                           yes -lgif
  Does Emacs use a png library?                           yes -lpng16 -lz 
  Does Emacs use -lrsvg-2?                                yes
  Does Emacs use cairo?                                   yes
  Does Emacs use -llcms2?                                 yes
  Does Emacs use imagemagick?                             no
  Does Emacs use native APIs for images?                  no
  Does Emacs support sound?                               yes
  Does Emacs use -lgpm?                                   yes
  Does Emacs use -ldbus?                                  yes
  Does Emacs use -lgconf?                                 no
  Does Emacs use GSettings?                               no
  Does Emacs use a file notification library?             yes -lglibc (inotify)
  Does Emacs use access control lists?                    yes -lacl
  Does Emacs use -lselinux?                               no
[...]

Issue #163, opened 2021-02-10
Each OS improvements


black7375:
### Linux
- Pure GTK support(https://deathwish.info/2020/08/01/pgtk-nativecomp/)
- Repo: https://github.com/flatwhatson/emacs#branch=pgtk-nativecomp
- Build Sample: https://github.com/mpsq/emacs-gcc-wayland-devel-builder

### Mac
- Already mentioned #99
- Repo: https://bitbucket.org/mituharu/emacs-mac/src/master/
- Build Sample: https://github.com/caldwell/build-emacs

### Windows
- Repo: https://ntemacs.sourceforge.io/

black7375:
@harryfei 
Hello?
Today, I briefly read the source of the emacs.
It's a small amount of time, so it may not be accurate..

What do you think about the coexistence of emacs-ng's webrender + [emacs:feature/pgtk](https://github.com/emacs-mirror/emacs/tree/feature/pgtk) || webrender + [emacs-mac](https://bitbucket.org/mituharu/emacs-mac/src/master/)?
Their processing methods are different and seem to overlap quite a lot.

- `emacs.c` (Webrender, PGTK, Mac)

https://github.com/emacs-ng/emacs-ng/blob/e7c4c721becd65fb1bc2507662880561c83eb30c/src/emacs.c#L1953-1956
https://github.com/emacs-mirror/emacs/blob/7a7bc15242896b20c7af49f77f0e22c3d78e4d88/src/emacs.c#L1929-1937
https://bitbucket.org/mituharu/emacs-mac/src/2345e4ca09f27f445adf12bc6613171dd4bfc5bf/src/emacs.c#lines-1957

- `xfaces.c`(Webrender, PGTK, Mac)

https://github.com/emacs-ng/emacs-ng/blob/e7c4c721becd65fb1bc2507662880561c83eb30c/src/xfaces.c#L574-592
https://github.com/emacs-mirror/emacs/blob/7a7bc15242896b20c7af49f77f0e22c3d78e4d88/src/xfaces.c#L578-596
https://bitbucket.org/mituharu/emacs-mac/src/2345e4ca09f27f445adf12bc6613171dd4bfc5bf/src/xfaces.c#lines-553

I just want to ask how likely are they.
From my point of view, it is mutually beta now, and even if possible, there will be a lot of work.

I'm curious about your opinion because you have experience in integrating webrender.

harryfei:
[...]

PR #185, opened 2021-03-17
Upstream feature/native-comp

black7375:
## Description
The build was successful and the environment is as follows:

```shell
## uname -a
# Linux black7375-pc 5.10.18-1-MANJARO #1 SMP PREEMPT Tue Feb 23 18:07:15 UTC 2021 x86_64 GNU/Linux

## rustc --version
# rustc 1.52.0-nightly (f5d8117c3 2021-03-16)

## clang --version
# clang version 11.1.0
# Target: x86_64-pc-linux-gnu

./autogen.sh
./configure --prefix=/usr --mandir=/usr/share/man --sysconfdir=/etc --libexecdir=/usr/lib --localstatedir=/var \
            --with-sound=alsa --with-modules --with-wide-int --with-xft --with-harfbuzz --with-cairo \
            --with-x-toolkit=gtk3 --with-xwidgets \
            --without-gconf --without-gsettings \
            --with-native-compilation
make -j"$(nproc)" CARGO_FLAGS="-j$(nproc)"
sudo make install
```

image
It runs in my environment.

## Warning
1. **Recreate `ChangeLog`**

I put `ChangeLog.3` back in because I plan to create a workflow that automatically merges upstream in the future.

2. **Webrender**

It doesn't work with option `webrender`, and it doesn't work with master in my environment.
```shell
--with-webrender
```
The `master` branch is the same.
More info: #184

3. **The most important thing**

There are many changes, so make sure to test them.

## TODO
I can't promise you I'll do it right away, but...
I want to do this kind of work.

- Creating a github action that automatically merges in [master branch](https://github.com/emacs-mirror/emacs/tree/master) and [feature/native-comp branch](https://github.com/emacs-mirror/emacs/tree/feature/native-comp)
- Build and release action
  - I'm thinking about it for AUR first.
- Combine [feature/pgtk branch](https://github.com/emacs-mirror/emacs/tree/feature/pgtk) for linux #163
  - It's not a common part, so I think we need a dedicated branch rather than a master. Like `target/linux`?


brotzeit:
Thanks for your work! It also works for me.
If you want an invite to the project, give this post a thumbs up.

> I put ChangeLog.3 back in because I plan to create a workflow that automatically merges upstream in the future.

What do you mean by automatically ? I'm asking because we still need the possibility to edit merges, since we did touch some C code and I suppose we will need to apply more modifications for features like webrender.

 > There are many changes, so make sure to test them.

I only tested some basic stuff since I heard that there have been issues with some packages recently. There's not much more we can do when merging upstream. I think it makes sense to tag certain commits where things seem to be stable, but this project will stay unstable and should be considered as playground for people who want to add new features and try new features. But I try to use it as my daily driver to see if I notice issues.

> Combine feature/pgtk branch for linux #163

    It's not a common part, so I think we need a dedicated branch rather than a master. Like target/linux?

Sure. I didn't take a look at the amount of changes they applied, but if we have a feature branch for some time and things seem to work, maybe we can also merge it. But maybe there are potential issues I don't see.

First thing when you intend to work on something, would ideally be to open an issue so we can discuss it.

brotzeit:
Let's give others a few days to test it, then I would merge it.

DavidDeSimone:
Concur with @brotzeit of letting this be tested for a few days and then merge.

black7375:
> What do you mean by automatically ?

This is an action that applies changes from the remote regularly (for example, on a weekly basis).
When this action is created, it will be updated automatically according to the schedule unless a conflict occurs.

> Sure. I didn't take a look at the amount of changes they applied, but if we have a feature branch for some time and things seem to work, maybe we can also merge it. But maybe there are potential issues I don't see.

 I think we need a new branch if the work is done because it can be unstable.

> First thing when you intend to work on something, would ideally be to open an issue so we can discuss it.

Yes, This PR was started simply to see if Upstream is possible, but if I do something else, I'll open the issue.

---

Thank you for the wonderful project!!

kiennq:
Btw, shouldn't in latest `nativecomp` branch, the build flag has been changed from `--with-nativecomp` to `--with-native-compilation`?

black7375:
I made a mistake by copying and putting the command in the issue #184.
Currently corrected it. The compilation itself is fine.

> Btw, shouldn't in latest nativecomp branch, 

Then, do we need a development branch, not a `master`?

> the build flag has been changed from `--with-nativecomp` to `--with-native-compilation`?

Yes. This is the latest branch option.

Should I change it to `--with-nativecomp`?
Or I can replace the document and test with `--with-native-compilation`.

black7375:
The invitation was well received.
Thank you. :smile_cat:

brotzeit:
Can we merge this ?

black7375:
There is still one consideration.(`--with-nativecomp` / `--with-native-compilation`)
https://github.com/emacs-ng/emacs-ng/blob/e7c4c721becd65fb1bc2507662880561c83eb30c/docs/faq.md#L13-L14

https://github.com/emacs-ng/emacs-ng/blob/7f4c75d249cf4a09ff69a2c80a84a3514ebd50d0/build_emacs_ng.sh#L12-L15

I think it's better to follow the latest options, but I haven't applied them because there's no agreement in thread yet.
> Should I change it to `--with-nativecomp`?
Or I can replace the document and test with `--with-native-compilation`.

kiennq:
I think we just update to use latest flag and it should be fine.

black7375:
Updated to latest emacs-ng commit to fix the build script.

Issue #186, opened 2021-03-18
Upgrade to Deno 1.8


DavidDeSimone:
Changelist is [here](https://deno.land/posts/v1.8)

I think there is a lot of good stuff in there for us, especially once deno-dropthrough is merged.

PR #190, opened 2021-03-21
Upgrading to Deno 1.8.1

DavidDeSimone:
Fixes https://github.com/emacs-ng/emacs-ng/issues/186

Changelog: https://deno.land/posts/v1.8


DavidDeSimone:
I'm running into what I think is a Cargo bug where for some reason, cargo cannot clone the wpt submodule of deno. What is very annoying is that we don't use that submodule to build deno. Removing that submodule from the .gitmodules doesn't help because, as far as I can tell, it will clone master and then update to our branch.

I think its because the .gitmodules file of deno has wpt be a shallow clone and cargo doesn't like that.

DavidDeSimone:
I fixed this by rolling back the commit that deno/test_utils/wpt pointed to be previous value from 1.7.2. Since wpt doesn't effect our runtime I don't expect this to be an issue. 

I still don't understand why cargo can't clone the current wpt hash. I'm seriously thinking it is a cargo bug that I might report to the cargo team.

If someone on OSX could try to build this it would be appreciated. I had to mess with some linker flags that seem to be working fine on ubuntu, but I would like to test on OSX. For now, tests all pass.

brotzeit:
Do we have any OSX users in the team ?

black7375:
My mac broke down about a year ago.

Can we reproduce it with Github Action or [Docker-OSX](https://github.com/sickcodes/Docker-OSX)?

DavidDeSimone:
Now that OSX builds are in CI, I'll merge in master to validate OSX builds properly. Once that is complete I'll merge this.

brotzeit (approved):

DavidDeSimone:
@brotzeit I don't think I understand why building deleted so much of the generated bindings. Is it because I didn't build with webrender enabled?

brotzeit:
> @brotzeit I don't think I understand why building deleted so much of the generated bindings. Is it because I didn't build with webrender enabled?

I ran into the same issue, but I'm actually not sure why this happens. Maybe we should just comment out the relevant code until we have fixed #203.

PR #192, opened 2021-03-23
Enable dynamic modules to directly read current buffer's text

ubolonton:
Since we want to be ABI-compatible with dynamic modules written for vanilla Emacs, we don't add more functions to `emacs_env`.

Instead, we provide a registry of named native functions that can be looked up at run time. We call these `ng-module functions`. Unlike normal module functions, they have globally stable addresses, which should be looked up once, at module load time (in `emacs_module_init`).

I made a [quick prototype](https://github.com/ubolonton/emacs-tree-sitter/commit/381cf5cdd300bf2df347f53f163d0d9d59635c36) that shows how to use this in `emacs-tree-sitter`. Some quick-n-dirty tests show up to 2x speedups.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6117,9 +6117,11 @@ esac
 if test "$HAVE_LIBVTERM" = "yes"; then
    CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"libvterm\", "
 fi
-
 if test "${with_javascript}" = "yes"; then
   CARGO_DEFAULT_FEATURES="${CARGO_DEFAULT_FEATURES}\"javascript\", "
+fi
+if test "${HAVE_MODULES}" = "yes"; then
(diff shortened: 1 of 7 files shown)


brotzeit:
Looks really cool! Can you add some sentences to our docs ? I would also link emacs-tree-sitter.

ubolonton:
Yeah, I'll write up a short doc with example.

DavidDeSimone (approved):
Code LGTM, agree it would be good to have some docs on this.

harryfei (approved):

DavidDeSimone (approved):
LGTM Code Wise

ubolonton:
I've added some docs for it.

The example looks a bit too long. Once `emacs-module-rs` adds support for this, it will be more concise.

brotzeit:
Nice!

Issue #195, opened 2021-03-29
error: implicit declaration of function 'unload_comp_unit'


nickali:
Hello, 

I had previously done a build a couple of weeks ago and it went fine. I did a pull again today and got the following error when I did a make:

```
alloc.c:3160:7: error: implicit declaration of function 'unload_comp_unit' is invalid in C99 [-Werror,-Wimplicit-function-declaration]

      unload_comp_unit (cu);
      ^
1 error generated.
make[1]: *** [alloc.o] Error 1
make: *** [src] Error 2
make: *** Waiting for unfinished jobs....
```
Sequence to do the build:

./autogen.sh
./configure
./make -j 4

I'm trying to do this on MacOS 11.2.3 and

```
❯ rustup --version
rustup 1.23.1 (3df2264a9 2020-11-30)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.51.0 (2fd73fabe 2021-03-23)`
```
and
```
❯ gcc --version
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin20.3.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

brotzeit:
Hi, thanks for the bug report. We have done an upstream merge recently and this seems related. Can you do me a favor and test the latest native-comp commit in emacs ? Unfortunately we have no OSX users in the community yet and we didn't add mac to our tests but we already have an issue for this I think.

black7375:
Maybe it's my PR.. #185
I only tested it on Linux because there was no Mac.

I will try to take compatibility measures.
First, I see if there are any changes upstream([emacs:master](https://github.com/emacs-mirror/emacs), [emacs:feature/native-comp](https://github.com/emacs-mirror/emacs/tree/feature/native-comp), [emacs:feature/native-comp-macos-fixes](https://github.com/emacs-mirror/emacs/tree/feature/native-comp-macos-fixes)).

Second, I will create a workflow that performs build tests.
The GitHub action allows you to test whether a build is possible with a Mac.
However, Windows build is not sure. #47

Finally, There are plans to investigate the compatibility of major packages.
[...]

PR #199, opened 2021-03-30
Upstream feature/native-comp for mac build

black7375:
Recent commit(79b8b6ca45ad707d86244882430e275efd95cdb9) resolves the Mac's build issue. #195


black7375:
Hmm...There seems to be still a problem.

https://github.com/black7375/emacs-ng/actions/runs/699932043
```
alloc.c:3160:7: error: implicit declaration of function 'unload_comp_unit' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
      unload_comp_unit (cu);
      ^
1 error generated.
make[1]: *** [alloc.o] Error 1
```

black7375:
I think we can include. src/comp.h
https://github.com/emacs-ng/emacs-ng/blob/bba450fab1e2a2ccc2dca40a6e5408596236d4a7/src/comp.h#L81

I recommend fixing this problem from the original.
Has anyone ever sent a patch to the FSF?

brotzeit:
Hopefully it's easier for us to fix this bug(and also to avoid this kind of issues in the future) when we have osx included in our build. I opened #200.

kidd:
native-comp got a fix for the warning: https://lists.gnu.org/archive/html/emacs-diffs/2021-04/msg00014.html

black7375:
I was about to go to bed, so I'll wake up tomorrow and merge it.

https://github.com/emacs-mirror/emacs/commit/978afd788fd0496540f715b83f18ed390ca8d5a4

Issue #196, opened 2021-03-29
Mac Build Test with native-comp enabled


black7375:
The Mac also performs build tests. #195

It will be in the form of adding to [.github/workflows/test.yml](https://github.com/emacs-ng/emacs-ng/blob/master/.github/workflows/test.yml)

declantsien:
Instead of updating [.github/workflows/test.yml](https://github.com/emacs-ng/emacs-ng/blob/master/.github/workflows/test.yml), since the macOS tests is already way slower than the Ubuntu ones.

Mac builds [here](https://github.com/emacs-ng/emacsng-darwin-builds) are with native-comp enabled.

Issue #58, opened 2021-01-05
emacs-ng has error with doom emacs install


jjpark78:
after succesfully build emacs-ng

run make install

and then 
i've copied nextstep/Emacs.app to Application folder 

and tryng to install doom emacs 
runs doom install 

got these errors

```
> ~/.emacs.d/bin/doom install
Debugger entered--Lisp error: (void-variable comp-ctxt)
  (type-of comp-ctxt)
  (memq (type-of comp-ctxt) cl-struct-comp-cstr-ctxt-tags)
  (and (memq (type-of comp-ctxt) cl-struct-comp-cstr-ctxt-tags) t)
  (progn (and (memq (type-of comp-ctxt) cl-struct-comp-cstr-ctxt-tags) t))
  (or (progn (and (memq (type-of comp-ctxt) cl-struct-comp-cstr-ctxt-tags) t)) (signal 'wrong-type-argument (list 'comp-cstr-ctxt comp-ctxt)))
  (progn (or (progn (and (memq (type-of comp-ctxt) cl-struct-comp-cstr-ctxt-tags) t)) (signal 'wrong-type-argument (list 'comp-cstr-ctxt comp-ctxt))) (aref comp-ctxt 4))
  (if range (progn (or (progn (and (memq (type-of comp-ctxt) cl-struct-comp-cstr-ctxt-tags) t)) (signal 'wrong-type-argument (list 'comp-cstr-ctxt comp-ctxt))) (aref comp-ctxt 4)) (progn (or (progn (and (memq (type-of comp-ctxt) cl-struct-comp-cstr-ctxt-tags) t)) (signal 'wrong-type-argument (list 'comp-cstr-ctxt comp-ctxt))) (aref comp-ctxt 3)))
[...]

PR #200, opened 2021-04-01
Include build for macOS

### LINKED ISSUE #36: configure github ci
Unfortunately travis has changed its payment model so we have to use github actions.

- [x] add seperate build for webrender
- [ ] Fix failing tests from native-comp branch(upstream)
- [x] Linux
- [x] OSX

brotzeit:
Close  #36 and #196

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,17 +4,30 @@ on: [push, pull_request]
 
 jobs:
     unix-test:
-        runs-on: ubuntu-latest
+        runs-on: ${{ matrix.os }}
+        strategy:
+            fail-fast: false
+            matrix:
+                os: [ubuntu-latest, macos-latest]
 
         steps:
             - uses: actions/checkout@v2
+
+            - name: Install depedencies on Linux
+              if: runner.os == 'Linux'
+              run: |
+                  sudo apt install -y build-essential texinfo libx11-dev libxpm-dev libjpeg-dev libpng-dev libgif-dev libtiff-dev libgtk2.0-dev libncurses-dev libxpm-dev automake autoconf libgnutls28-dev
+
+            - name: Install depedencies on macOS
+              if: runner.os == 'macOS'
+              run: |
+                brew install gnutls texinfo autoconf
 
             - name: rustfmt
               run: |
                   # Configure $PATH: Executables are installed to $HOME/bin
                   export PATH="$HOME/bin:$PATH"
 
-                  sudo apt install -y libgnutls28-dev
                   rustup install $(cat rust-toolchain)
                   rustup component add rustfmt-preview
                   ./autogen.sh && ./configure --without-makeinfo --with-x=no --with-ns=no --without-gconf --without-gsettings
@@ -22,7 +35,6 @@ jobs:
 
             - name: build
               run: |
-                  sudo apt install -y build-essential texinfo libx11-dev libxpm-dev libjpeg-dev libpng-dev libgif-dev libtiff-dev libgtk2.0-dev libncurses-dev libxpm-dev automake autoconf
                   ./autogen.sh && RUSTFLAGS="-Dwarnings" WERROR_CFLAGS='-Werror -Wno-error=deprecated-declarations'
                   ./configure --enable-checking=glyphs
                   make -j3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,15 +13,15 @@ jobs:
         steps:
             - uses: actions/checkout@v2
 
-            - name: Install depedencies on Linux
+            - name: Install dependencies on Linux
               if: runner.os == 'Linux'
               run: |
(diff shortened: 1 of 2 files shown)


brotzeit:
So at least we see the mac error in our build now =)

black7375:
I also did a build test a while ago and it is the same.
https://github.com/black7375/emacs-ng/blob/mac-build-test/.github/workflows/test.yml
https://github.com/black7375/emacs-ng/actions/runs/699906021

DavidDeSimone:
This is a great addition. I say we wait to merge until the OSX build issue is resolve, then I will merge https://github.com/emacs-ng/emacs-ng/pull/190 once I have confidence it won't break OSX.

ubolonton:
FWIW, I were never able to build the native-comp line with native compilation disabled. (It had issues with the dumper, and now mixing compile/run time checks (#195).) Building it with native compilation enabled *does work*, though.

I think we should encourage building with native compilation enabled. The CI should be built with that as well.

ubolonton:
Tangentially, maybe you guys already know this, but I find the [tmate action](https://github.com/mxschmitt/action-tmate/) to be pretty good for troubleshooting and incrementally working towards a working CI. It exposes an interactive shell into a GitHub Action runner. I [used it](https://github.com/ubolonton/tree-sitter-langs/blob/master/.github/workflows/shell.yml) for `tree-sitter-langs`.

DavidDeSimone:
@brotzeit what do you think of making the CI build with native comp enabled, and then turning native comp defaulted ON for emacs-ng?

brotzeit:
> Tangentially, maybe you guys already know this, but I find the tmate action to be pretty good for troubleshooting and incrementally working towards a working CI. It exposes an interactive shell into a GitHub Action runner. I used it for tree-sitter-langs.

Looks interesting, I'll give it a try.

> @brotzeit what do you think of making the CI build with native comp enabled, and then turning native comp defaulted ON for emacs-ng?

I think that's a good idea. There seems to be no good reason to try emacs-ng without native comp enabled.
Unless there are any objections, I would open a pull request with this change.

black7375:
The problem of `libgccjit` recognition failure in Mac is

> checking for gcc_jit_context_acquire in -lgccjit... no
configure: error: elisp native compiler requested but libgccjit not found.
Please try installing libgccjit or similar package.
If you are sure you want Emacs compiled without elisp native compiler, pass
  --without-native-compilation
to configure.
https://github.com/emacs-ng/emacs-ng/pull/200/checks?check_run_id=2259714672

GCC recompilation may be required.
> For that you need to compile gcc (duh). I edited Homebrew's gcc formula:
https://gist.github.com/mikroskeem/0a5c909c1880408adf732ceba6d3f9ab

brotzeit:
> GCC recompilation may be required.

Yeah I also found instructions that mention this. But I doubt it's a good idea to compile gcc in our build.

black7375:
I found an example that compiling is possible by using flags in the comments.
How about this one?

https://gist.github.com/mikroskeem/0a5c909c1880408adf732ceba6d3f9ab#gistcomment-3609148
```
export CC=clang
export LDFLAGS=-L/usr/local/Cellar/libgccjit/10.2.0_1/lib/gcc/10
export CXXFLAGS=-I/usr/local/Cellar/libgccjit/10.2.0_1/include
./configure --with-nativecomp
```

brotzeit:
Doesn't seem to work :(

black7375:
Hmm... it's difficult...

brotzeit:
I'm off for a few hours. You can push to this branch if you want.

DavidDeSimone:
I tried my hand at this based on what worked for a user back when this first came up for running doom-emacs on emacs-ng (https://github.com/emacs-ng/emacs-ng/issues/58) but it didn't work either.

Now that baseline OSX builds are fixed, maybe we just merge this and punt on enabling native-comp by default until we figure this out?

brotzeit:
Yeah, I also thought about that. It seems it's still not easy to get native comp working on mac, when looking at the tutorials. Maybe we are lucky and this changes soon. However we should mention the issue in our docs and recommend to try the tutorials. I will make the changes today and update our mac issue with this information.

.github/workflows/test.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
                   rustup install $(cat rust-toolchain)
                   rustup component add rustfmt-preview
-                  ./autogen.sh && ./configure --without-makeinfo --with-x=no --with-ns=no --without-gconf --without-gsettings
+                  ./autogen.sh && ./configure --without-makeinfo --with-x=no --with-ns=no --without-gconf --without-gsettings --with-native-compilation
```
DavidDeSimone, line 33: Should this be --without-native-compilation ?

black7375:
After remove native-comp flag, build able.
https://github.com/black7375/emacs-ng/actions/runs/718406142

.github/workflows/test.yml
brotzeit: I thought that it might make sense to build linux with native-comp enabled. But I can turn it off until it also works for macOS.

brotzeit: It was a little too late yesterday for me =)

DavidDeSimone (approved):
LGTM

Issue #204, opened 2021-04-06
Build fails on macOS due to variables from number.rs that are only defined for linux


brotzeit:
@DavidDeSimone I noticed that you added these `cfg` macros. The variables were duplicated so I removed them from remacs_sys. What was the reason that we had to do this ?

```rust
#[cfg(target_os = "linux")]
pub const INTMASK: EmacsInt = EMACS_INT_MAX >> (INTTYPEBITS - 1);
#[cfg(target_os = "linux")]
pub const MOST_POSITIVE_FIXNUM: EmacsInt = EMACS_INT_MAX >> INTTYPEBITS as u32;
#[cfg(target_os = "linux")]
pub const MOST_NEGATIVE_FIXNUM: EmacsInt = -1 - MOST_POSITIVE_FIXNUM;
```

Issue #197, opened 2021-03-29
[Meta] Compatibility of packages, distributions


black7375:
We can investigate compatibility with packages or distributions.

**Method**
It is investigated based on two.
- c961560317b5caaf19120e897a18b8c2fd08698a: before merge #185
- e1705229ca31ed450c5237573ba3f40c26207e07: merge

**Distributions**
- [ ] [Doom Emacs](https://github.com/hlissner/doom-emacs): #58
- [ ] [Spacemacs](https://github.com/syl20bnr/spacemacs)
- [ ] [Prelude](https://github.com/bbatsov/prelude)

**Packages**
Plan to list only incompatible packages.

brotzeit:
I wonder if it's easy to stay up-to-date with the compatibility notes that we put in our docs. We already had the situation that doom-emacs used a native-comp feature that we hadn't merged at this point.

black7375:
It's very easy to create upstream workflow.
How about creating a pull request once a month after automatically merging to `upstream` branch once a week?

brotzeit:
Sure, we can try that.

Issue #203, opened 2021-04-05
Don't generate bindings by default


brotzeit:
The bindings that are generated by bindgen are now included in the crate emacs #201. However they are still created every time we build emacs-ng. We need to put this code from src/Makefile.in into a phony target and make it optional to reduce the build time.

```
$(DEFINITIONS_FILE) $(BINDINGS_FILE) $(GLOBALS_FILE): $(filter-out ./macuvs.h,$(HEADERS)) $(rust_srcdir)/wrapper.h \
							$(rust_srcdir)/remacs-bindings/Cargo.toml \
							$(rust_srcdir)/remacs-bindings/src/main.rs
	$(MKDIR_P) $(dir $@)
	RUSTFLAGS="$(RUSTFLAGS)" \
	EMACS_CFLAGS="$(EMACS_CFLAGS)" \
	$(CARGO_RUN) --release --manifest-path=$(rust_srcdir)/remacs-bindings/Cargo.toml -- $(basename $(notdir $@)) $@
	touch $@
```

brotzeit:
Maybe it's better if we don't do it this way since we would have to include the bindings that are needed by each feature, even if emacs-ng is not compiled with all features. It also seems that the bindings are generated quite fast so I think it's probably not worth it.

declantsien:
This now implemented using [build.rs](https://github.com/emacs-ng/emacs-ng/blob/master/rust_src/crates/emacs/build.rs.in). I guess this issue is no long relevant.

PR #205, opened 2021-04-06
Efficient build CI

black7375:
There are a few more CI-related tasks to take place in the future. #196 #197
However, the slow build speed is hopeless.

Compared to the last successful build, the following improvements have been made:
https://github.com/emacs-ng/emacs-ng/actions/runs/720012839
https://github.com/black7375/emacs-ng/actions/runs/724927062

- Ubuntu: `26m 6s` => `10m 51s`
- MacOS: `39m 7s` => `16m 17s`

---

#204 be resolved first. (Mac Build Fail)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,12 +16,13 @@ jobs:
             - name: Install dependencies on Linux
               if: runner.os == 'Linux'
               run: |
-                  sudo apt install -y build-essential texinfo libx11-dev libxpm-dev libjpeg-dev libpng-dev libgif-dev libtiff-dev libgtk2.0-dev libncurses-dev libxpm-dev automake autoconf libgnutls28-dev zlib1g-dev libgccjit-9-dev
+                  sudo apt install -y build-essential texinfo libx11-dev libxpm-dev libjpeg-dev libpng-dev libgif-dev libtiff-dev libgtk2.0-dev libncurses-dev automake autoconf libgnutls28-dev zlib1g-dev libgccjit-9-dev
 
             - name: Install dependencies on macOS
               if: runner.os == 'macOS'
               run: |
                   brew install gnutls texinfo autoconf
+                  # libx11 libxpm jpeg libpng giflib libtiff gtk+3 ncurses libgccjit
 
             - name: rustfmt
               run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,9 +59,12 @@ jobs:
               run: |
                   export CARGO_INCREMENTAL=0
                   export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
-                  ./autogen.sh && RUSTFLAGS="-Dwarnings" WERROR_CFLAGS='-Werror -Wno-error=deprecated-declarations'
+                  ./autogen.sh && RUSTFLAGS="-Dwarnings" WERROR_CFLAGS='-Werror -Wno-error=deprecated-declarations' CFLAGS=' -g' CXXFLAGS=' -g'
                   ./configure --enable-checking=glyphs
-                  make -j$((`nproc`+1))
+
+                  # Avoid lisp compile && make info.
+                  # lib -> lib-src -> src -> lisp -> info
+                  make -j$((`nproc`+1)) src
 
             - name: js fmt
               run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,8 +58,13 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                os: [ubuntu-latest, macos-latest]
-                build: [general, webrender]
+                include:
+                    - os: ubuntu-latest
+                      build: general
+                    - os: ubuntu-latest
+                      build: webrender
+                    - os: macos-latest
+                      build: general
 
         steps:
             - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,6 @@ jobs:
               uses: actions/cache@v2
               with:
                   path: |
-                    ~/.cargo/registry/index
                     ~/.cargo/registry/cache
                     ~/.cargo/git
                     ./rust_src/target
@@ -60,11 +59,20 @@ jobs:
                   export CARGO_INCREMENTAL=0
                   export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
                   ./autogen.sh && RUSTFLAGS="-Dwarnings" WERROR_CFLAGS='-Werror -Wno-error=deprecated-declarations' CFLAGS=' -g' CXXFLAGS=' -g'
-                  ./configure --enable-checking=glyphs
+                  ./configure --enable-checking=glyphs --enable-rust-debug
 
                   # Avoid lisp compile && make info.
                   # lib -> lib-src -> src -> lisp -> info
                   make -j$((`nproc`+1)) src
+
+                  ## Reduce Cache
+                  # Note that all of the above need to be repeated for `release/` instead of
+                  # `debug/` if your build script builds artifacts in release mode.
+                  # https://bheisler.github.io/post/efficient-use-of-travis-ci-cache-for-rust/
+                  # Delete loose files in the debug directory
+                  find ./rust_src/target/debug -maxdepth 1 -type f -delete
+                  # Delete just meta data
+                  rm -f  ./rust_src/target/.rustc_info.json
 
             - name: js fmt
               run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,15 +17,21 @@ jobs:
               run: |
                   rustup install $(cat rust-toolchain)
                   rustup component add rustfmt-preview
+                  echo "::set-output name=week::$(/bin/date -u '+%Yy_%mm_%Ww')"
+                  echo "::set-output name=date::$(/bin/date)"
 
             - name: rust cache
               uses: actions/cache@v2
               with:
                   path: |
-                    ~/.cargo/registry/cache
-                    ~/.cargo/git
-                    ./rust_src/target
-                  key: ${{ runner.os }}-build-${{ hashFiles('**/rust-toolchain', '**/Cargo.toml*', '**/rustfmt.toml') }}
+                      ~/.cargo/registry/cache
+                      ~/.cargo/git
+                      ./rust_src/target
+                  key: ${{ runner.os }}-build-rust-${{ hashFiles('**/rust-toolchain', '**/Cargo.toml*', '**/rustfmt.toml') }}--${{ steps.rust-setup.outputs.week }}
+                  restore-keys: |
+                      ${{ runner.os }}-build-rust-${{ hashFiles('**/rust-toolchain', '**/Cargo.toml*', '**/rustfmt.toml') }}-${{ steps.rust-setup.outputs.week }}-test
+                      ${{ runner.os }}-build-rust-${{ hashFiles('**/rust-toolchain', '**/Cargo.toml*', '**/rustfmt.toml') }}-
+                      ${{ runner.os }}-build-rust
 
             - name: c cache
               uses: hendrikmuhs/ccache-action@v1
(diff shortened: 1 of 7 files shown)


brotzeit:
Nice, let's wait for #206. Hopefully we can reduce the build time even more. Disabling bindgen will also help a little.

black7375:
C, Lisp, and Typescript are possible to some extent , but unfortunately, I haven't learned Rust yet, so I'm not used to it. :cry: 
I plan to learn it at the end of the year, but it is not enough to understand rust related work or ecosystem.

brotzeit:
It's not about rust, but make #203. I already tried it in a recent pull request. Maybe somebody with more make experience can help out.

brotzeit:
Can you pleae rebase to get the mac fix from master.

black7375:
I am currently investigating linker.
Maybe today or tomorrow the results will come out.

brotzeit:
If you want, you can also add a seperate build for webrender. But I can also do it.

black7375:
Probably not that difficult.
I will try it with an update tomorrow.

black7375:
I'll update workflow based on this.

https://github.com/black7375/emacs-ng/actions/runs/731722328
https://github.com/black7375/emacs-ng/actions/runs/731774965

**Cold Build**
- Ubuntu: `17m 4s`
- Mac: `22m 44s`

**Hot Build**
- Ubuntu: `11m 53s`
- Mac: `14m 8s`

DavidDeSimone (approved):
LGTM

brotzeit:
@black7375 merge when you're ready.

black7375:
yeah

brotzeit:
Maybe we should rename the job "unix-test".

black7375:
There are some mistakes
It's moving, so I'll fix it a little later.

black7375:
> Maybe we should rename the job "unix-test".

How about `build-test`?

I'll wake up in the morning and simply write about the commit info.

---

There seems to be a problem with linux's webrender as well. https://github.com/emacs-ng/emacs-ng/issues/207#issuecomment-816814227

brotzeit:
> How about build-test?

Fine with me. I also can't think of anything better.

Out of curiosity, do you think it makes sense to put some parts into a seperate file or is this not possible without adding unnecessary overhead ?

black7375:
Excluding the cache-related part, it seems to be possible.
It's not that bad.

black7375:
### *NOTE ABOUT COMMIT*

The biggest bottlenecks were two.
- `cargo`: Downloading and compiling the Rust package takes a long time.
- `make`:  Only needed the executable `emacs`, but it was compiling all Lisp files.

I solved it like this:
- `cargo`: Caching `~/.cargo`, `rust_src/target`
- `make`: Omit `make lisp` and `make info`.

Regarding optimization options
- C: Most of the Lisp compilation is avoided, but still nearly half of the build time is spent compiling or generate Lisp.
  - At this point, the performance of the `C source` is important and you should use `-O2`.
  - So, if you build with -O0 or -O1: Fast for the C/Rust compilation, but slower for the later. As a result, it is slower. [Default -> O1: `52m 35s`](https://github.com/black7375/emacs-ng/actions/runs/727843913)
- Rust: It doesn't matter if you do it with `-O0`. Rather, optimizing it slows the build. [All O1 -> O2: `44m 23s`](https://github.com/black7375/emacs-ng/actions/runs/727846202)


---

### *Things we can try later in time*

I've focused on improving the main bottlenecks in this PR, and I'm not sure how fast the following items will be.

### Download
**Git Clone**
- needs: single-branch(https://github.com/actions/checkout/issues/346), sparse(https://github.com/actions/checkout/issues/172) support

**Docker**
- Article: [Caching Docker builds in GitHub Actions: Which approach is the fastest? 🤔 A research.](https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei)
- It may be faster if you download only the required environment using github container registry.

### Compiler
**[Zapcc](https://github.com/yrnkrn/zapcc)**: Caching C++ compiler based on clang, designed to perform faster compilations
- replace: `CC`, `CXX`, ccache
- needs: binary package, action/cache setting

**[rustc_codegen_cranlift](https://github.com/bjorn3/rustc_codegen_cranelift)**: Cranelift based backend for rustc
- article: [Using rustc_codegen_cranelift for debug builds](https://blog.rust-lang.org/inside-rust/2020/11/15/Using-rustc_codegen_cranelift.html)
- replace: rustc backend
- needs: rust's default support

### Linker
**[mold](https://github.com/rui314/mold)**: A modern Linker
- replace: `LD`, `OTHER_LDFLAGS`, `CCFLAGS`, `CXXFLAGS`, `RUSTFLAGS`

**[zld](https://github.com/michaeleisel/zld)**: A faster version of Apple's linker


### Build

**[Kati](https://github.com/google/kati)**: Makefile => Ninja
- needs: Kati does not support recursive make yet. https://github.com/google/kati/issues/165
- ubuntu: [ckati](https://packages.ubuntu.com/source/focal/android-platform-build-kati)

### Dependency
[Cargo Udeps](https://github.com/est31/cargo-udeps)
- result of command `cargo +nightly ude--all-targets`
```
unused dependencies:
`emacsng v0.1.0 (/home/black7375/Documents/emacs-ng/rust_src)`
└─── dependencies
     ├─── "itertools"
     ├─── "line-wrap"
     └─── "systemstat"
`lisp-macros v0.1.0 (/home/black7375/Documents/emacs-ng/rust_src/crates/lisp_macros)`
└─── dependencies
     └─── "lisp"
Note: They might be false-positive.
      For example, `cargo-udeps` cannot detect usage of crates that are only used in doc-tests.
      To ignore some dependencies, write `package.metadata.cargo-udeps.ignore` in Cargo.toml.
```

I haven't tried removing it yet, but you can try it later.

### Don't generate bindings
See #203

.github/workflows/test.yml
```diff
@@ -3,41 +3,137 @@ name: CI
...
+                  echo "::set-output name=date::$(/bin/date)"
+
+            - name: rust cache
+              uses: actions/cache@v2
+              with:
+                  path: |
+                      ~/.cargo/registry/cache
+                      ~/.cargo/git
+                      ./rust_src/target
+                  key: ${{ runner.os }}-${{ matrix.build }}-rust-${{ hashFiles('**/rust-toolchain', '**/Cargo.toml*', '**/rustfmt.toml') }}-${{ steps.rust-setup.outputs.week }}
```
kiennq, line 32: Why don't we use `${{ runner.os }}-cargo-${{ hashFiles('**/rust-toolchain', '**/Cargo.lock') }}` instead?
The `Cargo.toml` can be hand edited (`Cargo.lock` is better as key) and `rustfmt.toml` is just a format file.
Also the `-${{ steps.rust-setup.outputs.week }}` is not needed as we use `rust-toolchain already`

.github/workflows/test.yml
```diff
@@ -3,41 +3,137 @@ name: CI
...
+                  restore-keys: |
```
kiennq, line 33: Do we need the `restore-keys`? It seems that the cache is only stored using the key `${{ runner.os }}-${{ matrix.build }}-rust-${{ hashFiles('**/rust-toolchain', '**/Cargo.toml*', '**/rustfmt.toml') }}-${{ steps.rust-setup.outputs.week }}` that would not match the second/third entries.

.github/workflows/test.yml
```diff
@@ -3,41 +3,137 @@ name: CI
...
+
+            - name: configure(general)
+              if: ${{ matrix.build == 'general' }}
+              run: |
+                  ./configure --enable-checking=glyphs --enable-rust-debug
+
+            - name: configure(webrender)
+              if: ${{ matrix.build == 'webrender' }}
+              run: |
+                  ./configure --enable-checking=glyphs --enable-rust-debug --with-webrender
```
kiennq, line 121: Can we remove `--enable-rust-debug`? I'm trying to change the release pipeline to use the cache as well. However, the debug cache here is not working well with `release` that will be used for creating the release package.
Since it's cache, only the first build is slowed down right?

.github/workflows/test.yml
kiennq: Here is my PR that's using the proposed key, #216 
Hope that we can use the same key so the cache will be hit for both.

black7375: Don't we also define build options like codegen-unit in file `Cargo.toml`/`Cargo.toml.in`?
`-${{ steps.rust-setup.outputs.week }}`, `restore-keys updates` cache every week.
If features are added without changing dependencies, cache is broken.

.github/workflows/test.yml
black7375: If we use release mode, build will be slower because it uses optimizations like opt-level, lto.
I remember that it was slow even with cache.

.github/workflows/test.yml
black7375: As previously mentioned, this is a method for automatically updating the cache.
Even if the toolchain changes, we don't have to download it all packages from the beginning.

.github/workflows/test.yml
black7375: Build performance and optimization(runtime performance) are trade-offs, so I think it's better to separate them, but it's worth discussing.

.github/workflows/test.yml
kiennq: > Don't we also define build options like codegen-unit in file `Cargo.toml`/`Cargo.toml.in`?

That would be reflected in `Cargo.lock` wouldn't it? Which contains all of the dependencies packages the will be built.


> `-${{ steps.rust-setup.outputs.week }}`, `restore-keys updates` cache every week.
> If features are added without changing dependencies or fix build option, cache is broken.

I'm not sure if I get this correctly but you mean adding the `outputs.week` there is to invalidate the cache weekly?
Is that needed? Especially since the `restore-keys` just cut off that part brutally anyway.

.github/workflows/test.yml
```diff
@@ -3,41 +3,137 @@ name: CI
...
+            - name: rust cache
+              uses: actions/cache@v2
+              with:
+                  path: |
+                      ~/.cargo/registry/cache
+                      ~/.cargo/git
+                      ./rust_src/target
+                  key: ${{ runner.os }}-${{ matrix.build }}-rust-${{ hashFiles('**/rust-toolchain', '**/Cargo.toml*', '**/rustfmt.toml') }}-${{ steps.rust-setup.outputs.week }}
+                  restore-keys: |
+                      ${{ runner.os }}-${{ matrix.build }}-rust-${{ hashFiles('**/rust-toolchain', '**/Cargo.toml*', '**/rustfmt.toml') }}-${{ steps.rust-setup.outputs.week }}
```
kiennq, line 34: This first key is not needed since it has been set in `key`

.github/workflows/test.yml
black7375: > That would be reflected in `Cargo.lock` wouldn't it? Which contains all of the dependencies packages the will be built.

I couldn't find next options at lockfile.
```toml
[profile.dev]
panic = "abort"

[profile.release]
panic = "abort"
codegen-units = 1
lto = true
# This reduces the binary size from 125M to 70M due
# to Deno inclusion
opt-level = 'z' # Optimize for size
```
There is only dependency information.
After changing the profile option, considered hit rate.
I just thought defensively and I don't have a strong opinion as changes to file `Cargo.toml` are not frequent.

> I'm not sure if I get this correctly but you mean adding the `outputs.week` there is to invalidate the cache weekly?

It is not completely invalidated.
Build using old cache values, then upload updated cache(`1~2m`).
This is a policy to maintain freshness while making the most of the benefits of existing caches.

.github/workflows/test.yml
black7375: Fitst tested it, but can't find the cache key.
https://github.com/black7375/emacs-ng/runs/2320293270
```  
Cache not found for input keys: Linux-webrender-rust-a04d7bed436a11b5e00681f64766cdcfa2548c7b424f71095a6d22a6e3a66677-2021y_04m_15w, Linux-webrender-rust-a04d7bed436a11b5e00681f64766cdcfa2548c7b424f71095a6d22a6e3a66677-, Linux-webrender-rust
```

If you delete the key, an error will occur.
https://github.com/black7375/emacs-ng/runs/2320309690?check_suite_focus=true
```
Error: Input required and not supplied: key
```

Current policy seems to be correct.

.github/workflows/test.yml
kiennq: > There is only dependency information.
> After changing the profile option, considered hit rate.
> I just thought defensively and I don't have a strong opinion as changes to file `Cargo.toml` are not frequent.

Makes sense. However, for the caching stuff, this would be the same if you making a small change and rebuild in the `emacs-ng`, wouldn't it. So I expect the compiler should be able to detect that the cache of compiled binary is not usable in that case anyway.
The problem here is the cache of `~/.cargo/...`, which is just downloads of the prebuilt crate and source code. With the new cache key is created for changes in `Cargo.toml`, the cache there will be compressed, duplicated, and cached again, which's taking storage.



> It is not completely invalidated.
> Build using old cache values, then upload updated cache(`1~2m`).
> This is a policy to maintain freshness while making the most of the benefits of existing caches.

So after a week, a fresh cache will be created and taking more cache storage?
The updated cache is diff from the previous cache?

black7375: That's a good point.

**Compile Option Changes**
I intentionally reduced the debug information and compared the cache size.
https://github.com/black7375/emacs-ng/actions/runs/740067338
https://github.com/black7375/emacs-ng/actions/runs/740126744
- old cache: `~1762 MB`(Ubuntu General), `~1864 MB`(Ubuntu Webrender), `~1674 MB`(Mac General)
- new cache: `~1611 MB`(Ubuntu General), `~1689 MB`(Ubuntu Webrender), `~1629 MB`(Mac General)

**Lock File Changes**
After run `cargo update` to `Cargo.lock`
There may be slight increases.

- old cache: `~1762 MB`(Ubuntu General), `~1864 MB`(Ubuntu Webrender), `~1674 MB`(Mac General)
- new cache: ``(Ubuntu General), ``(Ubuntu Webrender), ``(Mac General)

...testing

black7375: But you can solve it like this.

https://github.com/matthiaskrgr/cargo-cache
`cargo cache --autoclean ` or `cargo cache --autoclean-expensive `

How about running after installing only when it is a new key?
It can be possible with `ouput.cache-hit`
```
if: steps.rust-cache.outputs.cache-hit != 'true'
```

black7375:
Will it merge after #207 is resolved?

brotzeit:
I'd prefer that we disable the webrender macOS build for now and note it in the issue that we have to enable it again after the problem is fixed.

brotzeit (approved):

brotzeit:
@black7375 you can merge when you're ready.

black7375:
@brotzeit  Unfortunately, I don't have merge permission.
Can you do it?

kiennq:
I've merged it

brotzeit:
@black7375 Now you should have write access.

black7375:
thanks!!

PR #206, opened 2021-04-06
define consts from number.rs for macOS too

### LINKED ISSUE #204: Build fails on macOS due to variables from number.rs that are only defined for linux

brotzeit:
close #204

### DIFF
--- a/rust_src/crates/emacs/src/number.rs
+++ b/rust_src/crates/emacs/src/number.rs
@@ -9,11 +9,8 @@ use crate::{
 
 // Largest and smallest numbers that can be represented as fixnums in
 // Emacs lisp.
-#[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


brotzeit:
I think this is good to go.

PR #210, opened 2021-04-10
ci: make release pipeline more efficient

kiennq:
Make the release pipeline more efficient by:
- Don't attempt to build a new release if the code has not changed from the last release
- Don't install `rust` in release pipeline anymore since it will be installed by the script later.
- Fix bug in build script that makes the `configure` not effective, making `native-compilation` Emacs not being built.
- Switch from `libgccjit-10` -> `libgccjit-9` for better compatible with older distro and also include more libs as deps

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,11 +17,12 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 2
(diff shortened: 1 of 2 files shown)


brotzeit (approved):
LGTM

DavidDeSimone (approved):
LGTM

PR #213, opened 2021-04-11
Nix: add locked rev and enhance develop env

GTrunSec:
Using locked rev for the saving binary cache. 
Adding  `nix-shell` of devshell to enhances the developer's environment who use Nix.

Adding cachix-key(send via email) to pushing the emacsNg binary cache by the maintainer

image


GTrunSec:
@ericdallo  https://github.com/numtide/devshell/blob/master/extra/language/rust.nix
I think we can add more options to devshell for the emacs-rust developers.

GTrunSec:
@ericdallo  https://github.com/numtide/devshell/blob/master/extra/language/rust.nix
I think we can add more options to devshell for the emacs-rust developers.

DavidDeSimone (approved):
LGTM

ericdallo (approved):
LGTM, nice!

IMO @GTrunSec @DavidDeSimone, we should recommend people who work on emacs-ng to dev/install it via nix, since it's pure and deterministic. They would just to install nix which is pretty simple. We would need to make that work for macOS still though

DavidDeSimone:
@ericdallo I haven't worked with nix before but I am eager to try it out. Is there a getting started guide or a tutorial for setup? My only concern is if we are going to encourage it for the primary dev env id like setup to be as straight forward as possible

ericdallo:
I totally agree @DavidDeSimone 
I think you can learn more with:
- https://nix-tutorial.gitlabpages.inria.fr/nix-tutorial/getting-started.html
- https://nixos.org/guides/how-nix-works.html

And if you want to take a deep dive:
https://nixos.org/guides/nix-pills/

GTrunSec:
> @ericdallo I haven't worked with nix before but I am eager to try it out. Is there a getting started guide or a tutorial for setup? My only concern is if we are going to encourage it for the primary dev env id like setup to be as straight forward as possible

yes, these features easy to use even the develop first time know it. the `getting started guide` doc is my next setup work.

GTrunSec:
> I think you can learn more with:

I have posted a [record](https://asciinema.org/a/O5gHLl8RV8sOFrogSdY3QWRsR). A simple tour of how to use it. 
The nix manual is too complicated, and you can’t spend less time trying to understand it. 
My purpose is to simply use and rapidly developed the emacsNg to reduce learning costs

ericdallo:
Tested manually and works great!

PR #214, opened 2021-04-11
add make target for bindgen

brotzeit:
Until we have found a solution for #203. I wasn't able to wrap this into a phony target. Maybe we have to change this section completely or I'm just missing something.


brotzeit:
Now bindgen can be triggered with `make bindgen`.

brotzeit:
I close this, maybe we decide differently in the future.

Issue #202, opened 2021-04-04
Add better javascript -> elisp integration


DavidDeSimone:
JavaScript has an automarshalling system that allows referencing of lisp variables from JS:

```js
const lispVar = lisp.list(1, 2, 3); 
lisp.print(lispVar)
```

However the other direction does not work the same

```lisp
(setq a (my-js-func))
(my-other-js-func a)
```

In the case where a is a reference to a JS variable, this doesn't work. The current work-around is to have `a` be a number that references a JS array index. We can do this automatically so it is transparent to the user.

DavidDeSimone:
I have a WIP for this, but as I play around with it, the more I think it needs to be opt in. Our current behavior is to JSON serialize an object, and then have lisp deser that object. The result is essentially a deep copy - which I prefer. The end result of that is by default JavaScript and Elisp prevent unintentional shared mutation by default. 

I think instead this behavior should be opt in, via 

```js
const x = lisp.make.proxy({x: 4});
return x;
```

PR #231, opened 2021-04-28
Adding reverse proxy logic

DavidDeSimone:
Fixes https://github.com/emacs-ng/emacs-ng/issues/202

~~Does not pass the "Deno Additions" JS text locally on my machine. Want to open this PR to show the interface to the world.~~

In the case that you want lisp to hold a reference to your javascript object, you can use the new function `lisp.make.proxy({...})`

Example:

```js 
lisp.defun({
    name: 'myfunc',
    func: () => {
	const biz = [];
	return lisp.make.proxy({ name: 'foo', bar: { biz: 'aooo' }, bb: biz });
    }
});

lisp.defun({
    name: "db",
    func: (db) => {
        console.log(db);
    }
});
```

```lisp
(setq x (myfunc))
(db x)
```

The reverse proxy is a specially formed cons that will clean up its javascript object once the lisp GC no longer holds a reference to it. 

@brotzeit request review


rust_src/src/javascript.rs
```diff
@@ -1514,6 +1615,14 @@ pub fn js__clear(idx: LispObject) -> LispObject {
     emacs::globals::Qnil
 }
 
+/// Internal function used for cleanup. Do not call directly.
+#[cfg(feature = "javascript")]
+#[lisp_fn]
+pub fn js__clear_r(idx: LispObject) -> LispObject {
+    execute_with_current_scope(move |scope| js_clear_r_internal(scope, idx));
+    emacs::globals::Qnil
```
brotzeit, line 1623: Drop this line and return nothing ?

rust_src/src/javascript.rs
```diff
@@ -763,6 +772,90 @@ pub fn is_proxy(
...
+        .unwrap()
+        .integer_value(scope)
+        .unwrap();
+
+    let finalizer = unsafe {
+        let mut bound = vec![
+            emacs::globals::Qjs__clear_r,
+            emacs::bindings::make_fixnum(idx.into()),
+        ];
+        let list = emacs::bindings::Flist(bound.len().try_into().unwrap(), bound.as_mut_ptr());
```
brotzeit, line 820: Maybe use `list!` ?

rust_src/src/javascript.rs
```diff
@@ -442,6 +442,15 @@ macro_rules! unproxy {
     }};
 }
 
+const REVERSE_PROXY_MARKER: i64 = 0x22fafd;
+
+macro_rules! make_reverse_proxy {
+    ($scope:expr, $lisp:expr) => {{
+        let marker = unsafe { emacs::bindings::make_int(REVERSE_PROXY_MARKER) };
```
brotzeit, line 449: LispObject::from_fixnum(REVERSE_PROXY_MARKER) ?

brotzeit: Oh I just realized there is a difference
```c
INLINE Lisp_Object
make_int (intmax_t n)
{
  return FIXNUM_OVERFLOW_P (n) ? make_bigint (n) : make_fixnum (n);
}
```
Do you think it makes sense to add a method for this ?

DavidDeSimone: I think this is worth addressing, but not needed for this PR. This implementation wouldn't support BIGNUMs on the JS side, so it's fine to just use the fixnum API since it will be simpler code.

DavidDeSimone:
Implemented your feedback @brotzeit 

Also instead of using a magic number, I will just use a generic symbol `js-proxy`. This makes the actual elisp proxies much more understandable.

brotzeit (approved):
LGTM

PR #234, opened 2021-05-01
Nix: add flake CI action

GTrunSec:


GTrunSec:
Too much time spent on the build. I think the building way could choose a binary cache if there have one.

Issue #236, opened 2021-05-01
Upgrade to Deno 1.9


DavidDeSimone:
Changelog: https://deno.com/blog/v1.9

Highlights: 
* LARGE performance increase by optimizing the deno "op" system. 
* Ability to swap permissions at runtime.

PR #237, opened 2021-05-02
Upgrading Deno to v1.9.1

DavidDeSimone:
Fixes https://github.com/emacs-ng/emacs-ng/issues/236

Changelog: https://deno.com/blog/v1.9

In addition to Deno, I'm also upgrading our rust version to current nightly.


DavidDeSimone:
These linker errors are strange to me - I am unable to reproduce them using clang/lld on my personal ubuntu machine. What's confusing is that the line numbers in the linker errors do not align with the actual code in https://github.com/denoland/rusty_v8/blob/v0.22.1/src/binding.cc

The OSX errors also seem to be related to the installation of rustfmt and not actual code.

DavidDeSimone:
I think this may be caused by caching the result of building v8/binding.cc ; what do you think @brotzeit ?

brotzeit:
Yeah looks like a caching issue again. However I'm still not sure how we can clear the cache easily. Maybe we should give this https://stackoverflow.com/a/64819132 a try.

DavidDeSimone:
@GTrunSec @ericdallo I'm confused on why nix isn't working. It looks like I've upgraded the Rust version to 04-25, but it's still giving me an error that is only on our previous nightly version. Any ideas why cargo/rustc don't seem to be properly updating?

GTrunSec (approved):

PR #239, opened 2021-05-03
webrender: add support for system clipboard.

harryfei:
Add support for system clipboard

Support both wayland and X11 clipboard API on Linux platform.


brotzeit:
I suppose this is also related to the cache ?

GTrunSec:
the CI CFALGS setting shoud be like this `CFLAGS=$CFLAGS -fuse-ld=lld -lxcb-render -lxcb-xfixes -lxcb-shape`

harryfei:
@GTrunSec wow, thanks.

src/frame.c
```diff
@@ -292,7 +292,7 @@ See also `frame-live-p'.  */)
     case output_ns:
       return Qns;
     case output_wr:
-      return Qwr;
+      return Qx;		/*  Pertend that we are X while actually wr */
```
brotzeit, line 295: Pertend -> Pretend

Issue #235, opened 2021-05-01
Prepare Dockerfiles to build emacs-ng


jman-schief:
The idea is to build a set of Dockerfiles to:
- Build emacs-ng in the CI
- Enable people to build emacs-ng themselves from scratch and without any context

Ideally building for the most common Linux distributions.

I'll start by cracking a Debian image.

jman-schief:
I think for now these files will suffice. In case we want to support a new distro we can always prepare a new Dockerfile image (or it can be contributed)

brotzeit:
Thanks for your work!

DavidDeSimone:
Good work, thank you.

PR #243, opened 2021-05-04
Add Dockerfile to build for Debian

jman-schief:
This will add a Dockefile to build a .deb for Debian.

It works but I'll iterate a bit more on this to see if I can slim this down a bit

(also suggestions and reviews welcome)

Ref: #235

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -1,0 +1,39 @@
+# Build emacs-ng using a Docker container
+
+To ease the job of building emacs-ng, we are preparing Docker images
+to allow a fully containerized build without installing any build dependencies.
+
+Bulding this will create a Docker image of about ~9 gb and will take
+quite some time (~45 minutes, depending on your workstation specs). Once the
+build is completed you can extract the artifact out of the container
+and install it.
+
+Usage example of the Dockerfile for building for Debian GNU/Linux:
+
+Clone the emacs-ng repository and build the image locally:
+``` sh
+$ git clone --depth=1 https://github.com/emacs-ng/emacs-ng.git
+$ cd emacs-ng
+$ docker build -t emacs-ng:builder -f docker/Dockerfile.debian .
+```
+
+At the end you should read this output:
+``` sh
+**********************************************************************
+
+ Done. The new package has been saved to
+
+ /src/emacs-ng_0.1-1_amd64.deb
+ You can install it in your system anytime using:
+
+      dpkg -i emacs-ng_0.1-1_amd64.deb
+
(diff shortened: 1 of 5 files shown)


docker/README.md
```diff
@@ -1,3 +1,28 @@
+# Build emacs-ng inside a Docker container
```
brotzeit, line 1: I think we should put the docs to the readme folder.

jman-schief: you mean moving this part into `./docs` (the web documntation), correct?

(I agree)

mkdocs.yml
```diff
@@ -10,6 +10,7 @@ nav:
     - Overview: README.md
     - Setup:
       - Build from source: building.md
+      - Build with Docker: docker.md
```
jman-schief, line 13: @brotzeit moved the relevant documentation here, is that ok?

jman-schief:
@brotzeit minus comments or corrections, I think we can merge this

black7375 (approved):

black7375:
Can we distribute image through [github docker registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry) or docker hub?

The disadvantage of the github registry is that [login is required](https://github.community/t/docker-pull-from-public-github-package-registry-fail-with-no-basic-auth-credentials-error/16358/87)  when downloading.

docs/docker.md
```diff
@@ -0,0 +1,39 @@
+# Build emacs-ng using a Docker container
+
+To ease the job of building emacs-ng, we are preparing Docker images
+to allow a fully containerized build without installing any build dependencies.
+
+Bulding this will create a Docker image of about ~9gb and will take
+quite some time (~45h, depending on your workstation specs). Once the
```
kidd, line 7: I hope it's 45m.... :)

jman-schief:
> Can we distribute image through [github docker registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry) or docker hub?

That would spare the user the time to build the image, but you will end up anyway with a 9gb (!) image locally. This Dockerfile is to ease the job of building emacs-ng, not to distribute it. For the purpose of distributing it, I would see more suitable a channel releasing binary packages (.deb, .rpm, .aur, ...), example a continuous delivery service with a github workflow.

black7375:
Ok, I understand.

DavidDeSimone (approved):
LGTM

mkdocs.yml
brotzeit: Yeah that's fine.

PR #252, opened 2021-05-07
Merge emacs master 2021-05-07

kidd:
only one conflict, in webrender part (which I can't test locally).

- the Ubuntu build works on my machine (TM) 
- Updated readme, 
- build script (added installing curl)
- gh workflows weren't `apt update` ing, and some dep failed due to trying to fetch a package not in the repo anymore. if my fix is too naive, say so and I'll see if there's some cache busting mechanism we can use.


rust_src/src/webrender_backend/term.rs
```diff
@@ -622,7 +622,6 @@ extern "C" fn read_input_event(terminal: *mut terminal, hold_quit: *mut input_ev
                     false,
                     true,
                     false,
-                    true,
```
brotzeit, line 625: @harryfei what do you think ?

harryfei: LGTM

PR #267, opened 2021-05-12
webrender: show different cursor types over different elements.

harryfei:
merge after PR #266


DavidDeSimone:
+ `if self.total_delta.y.abs() >= self.total_delta.y.abs()`

Should this be `if self.total_delta.x.abs() >= self.total_delta.y.abs()`?

harryfei:
@DavidDeSimone good catch. Thanks.

PR #268, opened 2021-05-13
Ignore Cargo.lock

brotzeit:
Since this file changes with different configure options, it seems to make sense to ignore this file.

### DIFF
--- a/rust_src/.gitignore
+++ b/rust_src/.gitignore
@@ -1,3 +1,5 @@
+Cargo.lock
+
 target
 generated/
 


DavidDeSimone:
We need to be careful with this - I use Cargo.lock to pin us to a
version of our Deno fork. Instead of doing that, I can have different
versions of the Deno fork be branches that I update Cargo.toml to use,
however that isn't how I have things set up right now.

declantsien:
Lockfiles are actually useful.
Cargo.lock shouldn't  change when using different configuration options.
Configure options are implemented using optional dependencies which are also pinned in lock file.

PR #275, opened 2021-05-17
add ng_async crate

### LINKED ISSUE #274: Move async code to seperate crate

brotzeit:
Close #274


brotzeit:
@DavidDeSimone  I removed some calls of

```
#[cfg(feature = "javascript")]
```

I ran into issues and thought it doesn't really

We need to apply some changes to the build script to be able to define `def_lisp_sym` under crates/. I'm not sure what needs to be done. I would like to move the code now and apply the necessary changes afterwards.

brotzeit:
I just noticed there's a difference between "Re-run jobs" and close and reopen a pull request. It just fixed an issue with ubuntu package installation.

rust_src/crates/ng_async/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+#![allow(non_camel_case_types, non_snake_case, non_upper_case_globals)]
+#![feature(concat_idents)]
+#![feature(async_closure)]
+
+extern crate libc;
+
+extern crate futures;
+extern crate lazy_static;
```
harryfei, line 8: I think we don't need most of those `extern crate`s in 2018 edition.

brotzeit: :+1:

DavidDeSimone:
I think this PR is good. Long term I think we will want to move
parsing into either it's own crate, or back into mainline. This is
only because what we really did was re-implement json.c in Rust, which
doesn't have to be used in an async context.

brotzeit:
> Long term I think we will want to move parsing into either it's own crate, or back into mainline.

The idea was(I think we agreed on this some time ago) to remove all code from rust_src/src, but sure we can add a seperate crate.

PR #278, opened 2021-05-18
Fix typo in autogen.sh

kamoii:
The rust relating code in autogen.sh refers to variable `remacs_version` which is not defined, causing output to look like:
```
$ ./autogen.sh                                                              
Checking Rust toolchain install ...
error: no default toolchain configured
error: no default toolchain configured
error: no default toolchain configured
Remacs currently requires Rust toolchain version .
Run 'rustup install '.
```

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -190,15 +190,15 @@ if [ -n $rust_toolchain_vers_path ] ; then
 
     case $retval in
         0) echo "Your system has the required Rust toolchain installed for building Remacs." ;;
-        1) echo >&2 "Remacs currently requires Rust toolchain version $remacs_version."
-	   echo >&2 "Run 'rustup install $remacs_version'."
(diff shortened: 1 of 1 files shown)


brotzeit:
Thanks!

autogen.sh
```diff
@@ -190,15 +190,15 @@ if [ -n $rust_toolchain_vers_path ] ; then
 
     case $retval in
         0) echo "Your system has the required Rust toolchain installed for building Remacs." ;;
```
brotzeit, line 192: Another remacs =)

brotzeit:
Sorry, but I think we should use "emacs-ng" as name.

DavidDeSimone (approved):
LGTM

black7375 (approved):

Issue #263, opened 2021-05-10
Building: emacs-ng seems to be unable to find system libs which is found by standard emacs


Delanii:
**Describe the bug**
Reposted from reddit per recommendetion. Original post: https://www.reddit.com/r/emacsng/comments/n8kiuq/building_from_source_emacsng_seems_to_be_unable/

emacs-ng seems to be unable to find system libraries which standard emacs is able to find and use in `.configure` results automatically.

Sample output of `./configure --with-native-compilation --with-webrender` on current emacs-ng source (part omitted for clarity):

```
  Does Emacs use -ljpeg?                                  no
  Does Emacs use -ltiff?                                  no
  Does Emacs use a gif library?                           no 
  Does Emacs use a png library?                           no 
  Does Emacs use -lrsvg-2?                                no
  Does Emacs use cairo?                                   no
  Does Emacs use -lgnutls?                                yes
  Does Emacs use -lxml2?                                  yes
  Does Emacs use -lfreetype?                              no
  Does Emacs use HarfBuzz?                                no
```

But if I am building standard emacs from source I get following output from `./configure --with-native-compilation` (again parts omitted):

```
  Does Emacs use -ljpeg?                                  yes
  Does Emacs use -ltiff?                                  yes
  Does Emacs use a gif library?                           yes -lgif
  Does Emacs use a png library?                           yes -lpng16 -lz
  Does Emacs use -lrsvg-2?                                yes
  Does Emacs use cairo?                                   yes
  Does Emacs use -lgnutls?                                yes
  Does Emacs use -lxml2?                                  yes
  Does Emacs use -lfreetype?                              yes
  Does Emacs use HarfBuzz?                                yes
```

I have tryed a few inclusion flags I could think of the get at least Harfbuzz recognized by emacs-ng, but none of them worked (but I never done that before, so I might doing that incorrectly).

According to reddit, this behaviour is not intended and should not even happen, but that is what I get from `./configure`. Could this be fixed on user side?

**To Reproduce**

```shell
git clone git://git.savannah.gnu.org/emacs.git
cd emacs
./autogen.sh
./configure --with-native-compilation
```

compare with:

```shell
git clone https://github.com/emacs-ng/emacs-ng
cd emacs-ng
./autogen.sh
./configure --with-native-compilation --with-webrender
```

**Expected behavior**
emacs-ng detects all appropriate dev headers and libraries in same manner as standard emacs. With that gives the same output of `.configure` script as standard emacs.

[...]

PR #284, opened 2021-05-23
webrender: support image rendering

harryfei:
Supports png, jpeg, git, tiff and pbm.

Re-implement image handlers in rust with [`image`](https://github.com/image-rs/image) crate instead of reusing the terrible codes in `GNU emacs`.

### DIFF
--- a/rust_src/src/webrender_backend/image.rs
+++ b/rust_src/src/webrender_backend/image.rs
@@ -14,8 +14,8 @@ use emacs::{
     lisp::LispObject,
 };
 use image::{
-    codecs::gif::GifDecoder, io::Reader, AnimationDecoder, DynamicImage, GenericImageView,
-    ImageFormat, ImageResult,
+    codecs::gif::GifDecoder, imageops::FilterType, io::Reader, AnimationDecoder, DynamicImage,
+    GenericImageView, ImageFormat, ImageResult,
 };
 use libc::c_void;
 use webrender::api::ImageKey;
@@ -24,6 +24,7 @@ use super::output::OutputRef;
 
 pub struct WrPixmap {
     pub image_key: ImageKey,
+    pub image_buffer: DynamicImage,
 }
 
 pub fn can_use_native_image_api(image_type: LispObject) -> bool {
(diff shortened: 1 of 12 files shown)


harryfei:
closes #284 #263

rust_src/src/webrender_backend/image.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    let lisp_data = match meta {
+        Some((frame_count, delay)) => {
+            let mut lisp_data = Qnil;
+
+            if frame_count > 0 {
+                lisp_data = LispObject::cons(
+                    Qcount,
+                    LispObject::cons(unsafe { make_fixnum(frame_count as i64) }, lisp_data),
```
brotzeit, line 126: Use `LispObject::from_fixnum`instead ?

rust_src/src/webrender_backend/image.rs
```diff
@@ -0,0 +1,230 @@
...
+        let image_key =
+            output.add_image(width, height, Arc::new(image_buffer.to_rgba8().into_raw()));
+
+        WrPixmap {
+            image_key,
+            image_buffer,
+        }
+    };
+
+    // take back old pixmap, let gc destory it's resource
```
brotzeit, line 161: destory it's -> destroy its

brotzeit (approved):

ReggieMarr:
@harryfei 
I just tested this pre-merge and it seems to work for me for pngs and some pdfs. 
That being said some pdfs don't seem scaled correctly and navigation is either slow or doesn't work.
Both these issues seem to stem from the same error `frame-edges: Symbol’s function definition is void: x-frame-edges`
I'm a little confused as to the architecture here. Are we not using xfns.c ? Is that supposed to be sort of replaced by frame.rs ? or is it replacing frame.el?
Additionally this seems to be x11 specific I assume your on wayland ?

harryfei:
`emacs` names some GUI api functions as `x` functions (such as `x_frame_edges`) for all the GUI backends(gtk, w32 and ns). We need to add the missing functions for `webrender` backend.

ReggieMarr:
Oh ok so probs we need to implement those functions in xfns which are used by say frame.el and window.el then eh?

PR #293, opened 2021-05-28
[WIP] Rename crate remacs-lib and delete unused code

### LINKED ISSUE #23: Remove unused code from remacs-lib
I think we should delete the folder completely since we don't seem to use the code. Or am I missing something ?

brotzeit:
In emacs-ng we only use this crate for `scan_rust_file`. Since the code has to be executed before bindgen, we have to keep the header file that was introduced in remacs. I would appreciate if anybody has a better idea for the new name.

close #23


DavidDeSimone:
I think you will need to change configure.ac to use the updated library archive name

brotzeit:
> I think you will need to change configure.ac to use the updated library archive name

Yeah I know, that was the point when I pushed my changes and stopped working =)

DavidDeSimone:
> > I think you will need to change configure.ac to use the updated library archive name
> 
> 
> 
> Yeah I know, that was the point when I pushed my changes and stopped working =)

Gotcha, sorry I missed the WIP in the title. 

I like the naming changes, and I think this is good work. 👍

declantsien:
`scan_rust_file` function is now in crate `lisp-doc`. Unused code has been removed in master.

PR #295, opened 2021-05-30
Add remacs docs

### LINKED ISSUE #223: Include relevant remacs docs in our wiki
I can imagine that some stuff from the remacs wiki could be worth adding to our docs. I'm pretty sure we have to modify it, but stuff like https://github.com/remacs/remacs/wiki/Emacs-Types would certainly be helpful for new contributors.

brotzeit:
close #223


brotzeit:
I didn't really review the docs, but I think most of the stuff also applies to emacs-ng.

jcs090218 (approved):

PR #299, opened 2021-06-03
bump libc to 0.2.95

brotzeit:


build-aux/config.guess
```diff
@@ -1483,8 +1482,8 @@ EOF
     i*86:rdos:*:*)
 	echo "$UNAME_MACHINE"-pc-rdos
 	exit ;;
-    i*86:AROS:*:*)
-	echo "$UNAME_MACHINE"-pc-aros
+    *:AROS:*:*)
+	echo "$UNAME_MACHINE"-unknown-aros
```
DavidDeSimone, line 1486: Why did we need to change this? Did it produce a new build error once we bumped libc?

harryfei:
is "0.2" enough to get latest version of crate?

build-aux/config.guess
brotzeit: This is not related to this pr, but this always changes when I build emacs-ng. Do these files not change when you run the build ?

brotzeit:
> is "0.2" enough to get latest version of crate?

No idea. But explicit seems more reasonable ?

build-aux/config.guess
brotzeit: I dropped the changes.

harryfei (approved):

DavidDeSimone:
I missed your response to my comment, but this looks like a good PR.
Glad it was merged.

Issue #57, opened 2021-01-04
offset_of! is unsound


harryfei:
https://github.com/remacs/remacs/issues/1564

PR #303, opened 2021-06-04
fix all warnings

harryfei:


rust_src/build.rs
```diff
@@ -497,7 +497,7 @@ fn build_ignored_crates(path: &PathBuf) -> bool {
 fn generate_crate_exports(path: &PathBuf) -> Result<(), BuildError> {
     let modules = find_crate_modules(&path.join("src"))?;
 
-    fs::create_dir(path.join("out"));
+    let _ = fs::create_dir(path.join("out"));
```
brotzeit, line 500: Do you think it makes sense to return `BuildError` when `crate_dir` fails ?

harryfei: It makes sense.
I try it. If the directory exists, this creation return error too.  That breaks our build process.

DavidDeSimone:
Does this new crate address https://github.com/emacs-ng/emacs-ng/issues/57, or is it's implantation of offset_of also unsound?

DavidDeSimone (approved):

harryfei:
@DavidDeSimone yes

PR #306, opened 2021-06-04
Webrender: fix known crash bugs

harryfei:


rust_src/crates/webrender/src/wrterm.rs
```diff
@@ -128,10 +130,64 @@ pub extern "C" fn get_keysym_name(keysym: i32) -> *mut libc::c_char {
...
+        if terminal.type_ != output_method::output_wr {
+            error!("Terminal {} is not a webrender display", terminal.id);
+        }
+
+        let dpyinfo = DisplayInfoRef::new(unsafe { terminal.display_info.wr as *mut _ });
+
+        return dpyinfo;
+    }
+
+    if obj.is_string() {
```
brotzeit, line 159: `if let Some(s) = obj.as_string()` ?

harryfei: Good idea

PR #309, opened 2021-06-05
Nix: Bumpup Src Version

GTrunSec:


GTrunSec:
Also, this rev for the Nix-darwin CI building as well. testshttps://github.com/GTrunSec/emacs-ng/runs/2750904179

PR #311, opened 2021-06-05
[RFC] use rust to implement main function

harryfei:
* Need platform compatibility test
* Collect ideas how can make this useful

### DIFF
--- a/rust_src/lib.rs
+++ b/rust_src/lib.rs
@@ -1,3 +1,27 @@
+use std::{os::raw::c_char, thread};
+
+extern "C" {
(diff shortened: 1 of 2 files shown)


harryfei:
Make emacs multiple threads?

brotzeit:
It certainly would be nice to apply such changes and improve emacs at this level. However I'm actually not sure if we are already at this point and thanks to David we can already run async code. Maybe there are good reasons to give this a try, but I think we should be very careful. I can only tell you that I'm not confident enough to merge stuff like this ;)

declantsien:
https://github.com/emacs-ng/emacs-ng/blob/master/rust_src/lib.rs

We already have this in the tree already. Closing this one.

PR #316, opened 2021-06-13
ignore build-aux/config.guess and build-aux/config.sub

brotzeit:
This really annoys me, but it seems I'm the only one with this issue ?


DavidDeSimone:
LTGM

PR #325, opened 2021-06-17
update handbook

brotzeit:


brotzeit:
@DavidDeSimone I've changed the readme a little bit so people don't think the project is only about js.

jcs090218 (approved):

brotzeit:
I'll merge this now. I would very much appreciate contributions to the handbook and also feedback on the readme layout. IMO the readme is very important to make a good first impression and therefore contain a good overview. On the other hand I want to avoid that users stop reading because they think the project is only about feature X. Maybe somebody has some good ideas how we can modify the readme to reflect the actual motivation of the project.

Issue #98, opened 2021-01-16
Decide which melpa packages should be included by default


yyoncho:
Current suggestions 

- use-package
- straight 

Potential candidates:

- magit 
- evil-mode
- projectile 
- flycheck
- company
- ivy + helm

Reference: melpa packages by downloads: https://melpa.org/#/?sort=downloads&asc=false

DavidDeSimone:
I think we should be careful on not including too much, but I think there is serious value in adding some additional base packages.

+1 for straight and magit

brotzeit:
@DavidDeSimone if we include straight we should also add use-package. I also agree on including magit. But IMO that's enough.

petr-tik:
My votes are for straight, magit and evil. 

since emacs-ng is based on emacs master/native-comp, it might be worth using core project.el instead of projectile (projectile is quite a heavy dependency). The main drawback might be the reliance of other packages on projectile, which may be alleviated with aliasing. 

Don't know if this is a good place to ask, please redirect me elsewhere if not. from my perusal of the code, it looks like some LSP client functionality is already integrated in rust. neovim's inclusion of a native LSP client is one of the reason its popularity took off so quickly imo.

brotzeit:
> it looks like some LSP client functionality is already integrated in rust

Yeah that was one of the initial goals. It's currently just stuck at the moment. Maybe we should add an issue to track the progress.

brotzeit:
> My votes are for straight, magit and evil.

evil is too opionated for my tase.

jcs090218:
> since emacs-ng is based on emacs master/native-comp, it might be worth using core project.el instead of projectile (projectile is quite a heavy dependency). The main drawback might be the reliance of other packages on projectile, which may be alleviated with aliasing.

I agree with @petr-tik, I have used both of these packages and `project.el` is indeed faster and a lot more lightweight.

`use-package` may be an good option.

`flycheck` seems to be better than `flymake` so I think it would be great to have it! :)

I think `helm` is better for people who just started use Emacs. I personally like `ivy` since it's faster and more lightweight once people are getting better at elisp.

---

My votes are `flycheck`, `use-package`, `ivy`, `project.el`.

brotzeit:
I would like to not force too many packages on the user especially when there are widely used alternatives.

kidd:
I gave a shot to that, and there are a couple of things I wanted to check with you.

[...]

PR #346, opened 2021-07-11
add straight.el/use-package as git submodules

declanqian:
#98

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,8 @@
+[submodule "use-package"]
+	path = lisp/use-package
+	url = https://github.com/jwiegley/use-package.git
+	shallow = true
+[submodule "straight.el"]
+	path = lisp/straight
+	url = https://github.com/raxod502/straight.el.git
+	shallow = true
--- a/autogen.sh
+++ b/autogen.sh
@@ -456,6 +456,14 @@ if test -n "$tailored_hooks$sample_hooks"; then
     fi
 fi
 
+if $do_git; then
+    echo "Updating git submodules..."
+    git submodule update --init --recursive
+    git submodule update --recursive
+else
+    git_was_ok=false
+fi
+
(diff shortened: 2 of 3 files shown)


brotzeit:
Thanks for the pull request! Please give us some time to review your changes.

FYI I just looked if there's progress on merging use-package into emacs, but it seems nothing happened. @declanqian we should keep the related issue open as a reminder that we have to remove use-package when it was merged upstream.

declanqian:
Sure, no problem. Take your time.

> we should keep the related issue open 
as a reminder that we have to remove use-package when it was merged upstream.

That does make sense. Let's do that.

DavidDeSimone:
What is the origin of the changes in ldef-boot.el? Is this an older version behind upstream for compatibility?

declanqian:
It's a copy of the generated `loaddefs.el` from lisp dir. This file is only
used in the early stage of the build process I think. Should be OK to leave
it out though. @DavidDeSimone

declanqian:
It should an output of `./admin/update_autogen`.
Since we're using git submodule, the script is broken.
Anyway, just left that changes out.

DavidDeSimone:
Thanks for the PR, this looks good to me. I will defer to the team members that have a stronger knowledge of these packages but it's a 👍 for me.

kidd (approved):
Triggered the tests. looks good to me

brotzeit:
I think we should mention this somewhere in the docs, any ideas ?

black7375:
https://emacs-ng.github.io/emacs-ng/#features
How about adding a built-in packages?

brotzeit:
> How about adding a built-in packages?

Maybe it's better to add a section under "Home" since all the other features are based on rust.

I'll open a seperate doc issue.

brotzeit:
@declanqian Thanks!

PR #350, opened 2021-07-17
webrender: support font fallback.

harryfei:


rust_src/crates/webrender/src/font_db.rs
```diff
@@ -0,0 +1,97 @@
...
+            family_serif: Self::default_font_family(FamilyName::Serif),
+            family_sans_serif: Self::default_font_family(FamilyName::SansSerif),
+            family_cursive: Self::default_font_family(FamilyName::Cursive),
+            family_fantasy: Self::default_font_family(FamilyName::Fantasy),
+            family_monospace: Self::default_font_family(FamilyName::Monospace),
+        }
+    }
+
+    pub fn select_family(&self, family: &Family) -> Vec<&FaceInfo> {
+        let family = self.family_name(family);
```
brotzeit, line 35: Can we use `if let Some(f)` so we can drop the `unwrap` below ?

brotzeit (approved):

Issue #331, opened 2021-06-19
webrender: Emacs window missing `app_id` under Wayland swaywm


declanqian:
Below is the relevant part from output of `swaymsg -t get_tree`

```
{
  "id": 210,
  "type": "con",
  "orientation": "none",
  "percent": 0.5,
  "urgent": false,
  "marks": [
  ],
  "focused": false,
  "layout": "none",
  "border": "pixel",
  "current_border_width": 0,
  "rect": {
    "x": 960,
    "y": 24,
    "width": 960,
    "height": 1056
  },
  "deco_rect": {
    "x": 0,
    "y": 0,
    "width": 0,
    "height": 0
  },
  "window_rect": {
    "x": 0,
    "y": 0,
    "width": 960,
    "height": 1056
  },
  "geometry": {
    "x": 0,
    "y": 0,
    "width": 800,
    "height": 600
  },
  "name": "*scratch* - GNU Emacs at thinkpadx1extreme",
  "window": null,
  "nodes": [
  ],
  "floating_nodes": [
  ],
  "focus": [
  ],
  "fullscreen_mode": 0,
  "sticky": false,
  "pid": 988648,
  "app_id": null,
  "visible": true,
  "max_render_time": 0,
  "shell": "xdg_shell",
  "inhibit_idle": false,
  "idle_inhibitors": {
    "user": "none",
    "application": "none"
  }
}
```

declanqian:
After dive in a little, figured out this seems like unimplemented chore for [glutin on Wayland](https://docs.rs/glutin/0.27.0/glutin/platform/unix/trait.WindowBuilderExtUnix.html#tymethod.with_app_id).

With a quick web search, I was able to find [`Alacritty`'s implementation](https://github.com/alacritty/alacritty/blob/master/alacritty/src/display/window.rs#L301)

The reason I fired this issue in the first place is that WR frames are not respecting my Sway 
shortcut for closing a window. I thought it has something to do with the `app_id`, 
Which seems now to be an unimplemented chore, too.
In this case, [`glutin::event::WindowEvent::CloseRequested`](https://github.com/rust-windowing/glutin/blob/master/glutin_examples/examples/window.rs#L28) maybe?

declanqian:
> WR frames are not respecting my Sway shortcut for closing a window

This now been tracked from #358

PR #355, opened 2021-07-18
webrender: set app_id for wayland

declanqian:
#331


harryfei (approved):

declanqian:
I followed [`Alacritty`'s implementation](https://github.com/alacritty/alacritty/blob/master/alacritty/src/display/window.rs#L301).
As far as I can tell, PGTK Emacs doesn't do this explicitly, GTK does it for us by using the executable's name. So I did the same for WR Emacs.

Issue #348, opened 2021-07-14
Document built-in packages use-package and straight


brotzeit:
Follow up issue of #346

brotzeit:
I think we have to apply some modifications to make this change useful for users.

This is my straight config:

```
(setq byte-compile-warnings '(cl-functions))
(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "~/home/straight.el/bootstrap.el" "~/.newemacs.d/"))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

;;; https://github.com/raxod502/straight.el/blob/develop/README.md#integration-with-use-package
(straight-use-package 'use-package)
(setq straight-use-package-by-default t)
```

My config is quite old so I don't know if there's a better way to configure straight, but we have to define the location of the built-in straight.el package. It also seems it's necessary to add use-package as default method.

Any straight users here with an idea ?

declanqian:
I am using this
```
(if (fboundp 'straight-use-recipes)
    (progn
      // copy some code from bootstrap.el
      )
  (let ((bootstrap-file
         (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
        (bootstrap-version 5))
    (unless (file-exists-p bootstrap-file)
      (with-current-buffer
          (url-retrieve-synchronously
           "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
           'silent 'inhibit-cookies)
        (goto-char (point-max))
        (eval-print-last-sexp)))
    (load bootstrap-file nil 'nomessage)))  
```

I think basically what we need is just to setup some recipes
using `straight-use-recipes` and turn on some minor modes `straight-use-package-mode` etc.
I sent a pull request before to wrap this in a function called `straight-bootstrap`,
later deleted it by thinking it's not worth it.

I liked `package.el`'s `package-initialize` logic, but it can complicate thing up.

brotzeit:
We could add the bootstrapping code to `/home/fredbert/ng_stuff/emacs-ng/lisp/emacs-lisp/emacs-ng.el`.

declanqian:
[...]

PR #365, opened 2021-08-01
Automate straight.el bootstrapping process

declanqian:
Context https://github.com/emacs-ng/emacs-ng/issues/348#issuecomment-890514436

### DIFF
--- a/lisp/emacs-lisp/emacs-ng.el
+++ b/lisp/emacs-lisp/emacs-ng.el
@@ -1,1 +1,110 @@
+;;; emacs-ng.el --- Emacs NG related -*- lexical-binding: t -*-
+
+;; `straight.el'
+(defvar straight-recipes-gnu-elpa-use-mirror)
+(defvar straight-recipes-emacsmirror-use-mirror)
+(defvar straight-use-symlinks)
+(defvar straight-enable-package-integration)
+(defvar straight-enable-use-package-integration)
+(defvar straight-recipe-repositories)
+(defvar straight-recipes-gnu-elpa-url)
+(defvar straight-repository-user)
+(defvar straight-repository-branch)
+(declare-function straight--reset-caches "straight")
+(declare-function straight-use-package-mode "straight")
+(declare-function straight-package-neutering-mode "straight")
+(declare-function straight-symlink-emulation-mode "straight")
+(declare-function straight-watcher-start "straight")
+(declare-function straight-live-modifications-mode "straight")
+(declare-function straight--modifications "straight")
+
+;;; Customization options
+
+;;;###autoload
+(defcustom ng-straight-bootstrap-at-startup nil
+  "Whether to bootstrap straight.el when Emacs starts.
+If non-nil, straight.el is bootstrapped before reading the init
+file (but after reading the early init file).  This means that if
+you wish to set this variable, you must do so in the early init
+file.  Regardless of the value of this variable, straight.el recipes
+are not made available if `user-init-file' is nil (e.g. Emacs
+was started with \"-q\").
+
+Even if the value is nil, you can type \\[ng-bootstrap-straight] to
+bootstrap straight.el at any time, or you can
+call (ng-bootstrap-straight) in your init-file."
+  :type 'boolean
+  :group 'emacs-ng
+  :version "28")
+
+;;;###autoload
+(defvar ng--straight-bootstrapped nil
+  "Non-nil if `ng-bootstrap-straight' has been run.")
+
+;;;###autoload
+(defun ng-bootstrap-straight ()
+  "Bootstrap straight.el."
+
+  (interactive)
+
+  (require 'straight)
+
+  (setq ng--straight-bootstrapped t)
+
+  (straight--reset-caches)
+  (setq straight-recipe-repositories nil)
+
+  (straight-use-recipes '(org-elpa :local-repo nil))
+
+  (straight-use-recipes '(melpa :type git :host github
+                                :repo "melpa/melpa"
+                                :build nil))
+
+  (if straight-recipes-gnu-elpa-use-mirror
+      (straight-use-recipes
+       '(gnu-elpa-mirror :type git :host github
+                         :repo "emacs-straight/gnu-elpa-mirror"
+                         :build nil))
+    (straight-use-recipes `(gnu-elpa :type git
+                                     :repo ,straight-recipes-gnu-elpa-url
+                                     :local-repo "elpa"
+                                     :build nil)))
+
+  (straight-use-recipes '(el-get :type git :host github
+                                 :repo "dimitri/el-get"
+                                 :build nil))
+
+  (if straight-recipes-emacsmirror-use-mirror
+      (straight-use-recipes
+       '(emacsmirror-mirror :type git :host github
+                            :repo "emacs-straight/emacsmirror-mirror"
+                            :build nil))
+    (straight-use-recipes '(emacsmirror :type git :host github
+                                        :repo "emacsmirror/epkgs"
+                                        :nonrecursive t
+                                        :build nil)))
+
+  ;; Prefer newly included straight and use-package
+  (straight-override-recipe `(straight :type built-in))
+  (straight-override-recipe `(use-package :type built-in))
+
+  (if (straight--modifications 'check-on-save)
+      (straight-live-modifications-mode +1)
+    (straight-live-modifications-mode -1))
+
+  (when (straight--modifications 'watch-files)
+    (straight-watcher-start))
+
+  (if straight-use-symlinks
+      (straight-symlink-emulation-mode -1)
+    (straight-symlink-emulation-mode +1))
+
+  (if straight-enable-package-integration
+      (straight-package-neutering-mode +1)
+    (straight-package-neutering-mode -1))
+
+  (if straight-enable-use-package-integration
+      (straight-use-package-mode +1)
+    (straight-use-package-mode -1)))
+
 (provide 'emacs-ng)
--- a/Makefile.in
+++ b/Makefile.in
@@ -873,7 +873,7 @@ uninstall-nt:
 
 # ==================== Cleaning up and miscellanea ====================
 
-.PHONY: mostlyclean clean distclean bootstrap-clean maintainer-clean extraclean
+.PHONY: mostlyclean clean distclean bootstrap-clean maintainer-clean extraclean submodule
 
 ## Eg:
 ## src_clean:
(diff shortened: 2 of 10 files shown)


lisp/emacs-lisp/emacs-ng.el
```diff
@@ -1 +1,112 @@
...
+      (straight-use-recipes
+       '(emacsmirror-mirror :type git :host github
+                            :repo "emacs-straight/emacsmirror-mirror"
+                            :build nil))
+    (straight-use-recipes '(emacsmirror :type git :host github
+                                        :repo "emacsmirror/epkgs"
+                                        :nonrecursive t
+                                        :build nil)))
+
+  (straight-use-package `(straight :type git :host github
```
declanqian, line 87: Should we left this out?

declanqian:
The function names can be improved I think.

brotzeit:
Sorry for the delay, I wasn't able to build it successfully. However, I just ran git clean and now it seems to be working fine.

But I always get dirty submodules. Do you think it makes sense to ignore it with `ignore = dirty` in the submodule config ? Or does this have any side effects we don't want ?

lisp/emacs-lisp/emacs-ng.el
brotzeit: Since we use a submodule for straight, I guess we don't need this.

brotzeit:
Can you please add some notes to https://github.com/emacs-ng/emacs-ng/blob/master/mkdocs.yml#L9.

Just a new section for "use-package and straight" below "Setup". We can improve the docs later, but users should know about it.

declanqian:
> But I always get dirty submodules. Do you think it makes sense to ignore it with ignore = dirty in the submodule config ? Or does this have any side effects we don't want ?

`ignore = dirty` doesn't help in this case. I've added a make `submodule` command to update
submodules. Also, submodules are updated after `make clean` and `./autogen.sh`.
You can use `magit-submodule` to update a individual submodule.

Added some notes in `./admin/notes/git-submodule`.

> Can you please add some notes to

Added a subheading `Built-in Emacs Lisp packages` under `Features` in `README.md` file

> The function names can be improved I think.

Rename `ng/` to `ng-` according to [Emacs Lisp Coding Conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html)
I've used `ng` as the short word to distinguish `emacs-ng` from other Lisp programs

brotzeit:
@declanqian sorry again for the delay, it's too hot for me =)

I still had some issues and then I realized that straight doesn't respect a different value for `user-emacs-directory`. There seems to be code in install.el that's necessary.

I think we have to tell the user to put an additional line into his config before installing packages. Or do you see a possibility how we can avoid it ?

declanqian:
Oh, you must do the additional straight lines in the early init file, as the straight bootstrap is done before
loading the regular init file. Just like the way `package-enable-at-startup` works.
@brotzeit 
I have these in my early-init.el file, and it works fine.
```
(setq package-enable-at-startup nil)
(setq straight-repository-branch "develop"
      straight-base-dir (concat user-emacs-directory "var/")
      straight-enable-use-package-integration t
      straight-use-package-by-default t
      straight-build-dir (format
                          "build-%s__%s"
                          emacs-version
                          (number-to-string
                           (time-to-seconds emacs-build-time)))
      straight-cache-autoloads nil
      straight-check-for-modifications nil
      straight-enable-package-integration nil
      straight-vc-git-default-clone-depth 1
      autoload-compute-prefixes nil)
```

declanqian:
Nah, after a second thought. Straight.el only use `user-emacs-directory` as
`straight-base-dir`. If you change `user-emacs-directory` from the `early-init.el`,
`straight.el` should respect you settings.

You also can set `ng-straight-bootstrap-at-startup` to `nil` from early init file, then
manully invoke `ng-bootstrap-straight` afterwards.

`straight.el`'s `bootstrap.el` and `install.el` files have no use here. Probably need a
way to remove them from the final distribution.

brotzeit:
> straight.el's bootstrap.el and install.el files have no use here. Probably need a
way to remove them from the final distribution.

I think we can ignore it. And we can add documentation on how to change `emacs-user-directory` later.

But I still got an issue with loading themes:

```
Debugger entered--Lisp error: (error "Invalid face attribute name" :files)
  internal-set-lisp-face-attribute(default :files ("*.el" ("recipes" "recipes/el-get.rcp") "methods" "el-get-pkg.el") #<frame  *Minibuf-1* - GNU Emacs at brotzeit 0x564c01bd67f0>)
  set-face-attribute(default #<frame  *Minibuf-1* - GNU Emacs at brotzeit 0x564c01bd67f0> :underline nil :overline nil :strike-through nil :box nil :inverse-video nil :stipple nil :inherit nil :extend nil :files ("*.el" ("recipes" "recipes/el-get.rcp") "methods" "el-get-pkg.el") :flavor melpa)
  apply(set-face-attribute default #<frame  *Minibuf-1* - GNU Emacs at brotzeit 0x564c01bd67f0> (:underline nil :overline nil :strike-through nil :box nil :inverse-video nil :stipple nil :inherit nil :extend nil :files ("*.el" ("recipes" "recipes/el-get.rcp") "methods" "el-get-pkg.el") :flavor melpa))
  face-spec-reset-face(default #<frame  *Minibuf-1* - GNU Emacs at brotzeit 0x564c01bd67f0>)
  face-spec-recalc(default #<frame  *Minibuf-1* - GNU Emacs at brotzeit 0x564c01bd67f0>)
  custom-theme-recalc-face(default)
  enable-theme(wombat)
  load-theme(wombat nil nil)
```

When I remove

```
  (straight-use-recipes '(el-get :type git :host github
                                 :repo "dimitri/el-get"
                                 :build nil))
```

everything works fine.

brotzeit:
@declanqian I've tested it again and everything seems to be working. And changing the emacs directory also works with the early init file. I'm actually not even sure I've heard of this before.

declanqian:
Yeah, it happens if we load `emacs-lisp/emacs-ng` from very early stage `loadup`.
It corrupts the elisp `default` symbol somehow. So I've removed that.

Glad things work out for you. I've done changing this branch BTW. @brotzeit

brotzeit (approved):

brotzeit:
Forgot to approve...

brotzeit:
Maybe rebase on master so tests go green ?

harryfei (changes requested):
If conflict with doom-emacs
```
Warning (straight): Two different recipes given for "straight" (:type cannot be both built-in and git) Disable showing Disable logging
```

brotzeit:
> It conflicts with doom-emacs

I guess we have to fix this in doom-emacs right ?

declanqian:
The warning seems can be emliated with `straight-override-recipe`


``` diff
diff --git a/lisp/emacs-lisp/emacs-ng.el b/lisp/emacs-lisp/emacs-ng.el
index 8d0d8e550e..a2e6f0cb33 100644
--- a/lisp/emacs-lisp/emacs-ng.el
+++ b/lisp/emacs-lisp/emacs-ng.el
@@ -84,8 +84,9 @@ ng-bootstrap-straight
                                         :nonrecursive t
                                         :build nil)))

-  (straight-use-package `(straight :type built-in))
-  (straight-use-package `(use-package :type built-in))
+  ;; Prefer newly included straight and use-package
+  (straight-override-recipe `(straight :type built-in))
+  (straight-override-recipe `(use-package :type built-in))

   (if (straight--modifications 'check-on-save)
       (straight-live-modifications-mode +1)
```


But to be clear, doom is not using the built-in straight.el.

brotzeit:
But we should then probably add a note ?

declanqian:
I think [notes](https://github.com/emacs-ng/emacs-ng/blob/002f7deba9e1494dfbcd1ed7c819e13f0c341550/README.md#usage) here are covered.
For doom distribution to take advantage of the newly builtin straight and use-package,
I can try sent a PR, not sure whether they will accept that.

harryfei:
doom-emacs user could choose to disable `ng-straight-bootstrap-at-startup`. Am I right?

harryfei (approved):

declanqian:
> doom-emacs user could choose to disable `ng-straight-bootstrap-at-startup`. Am I right?
> Can we add this in our doc too?

There are some occurs in the README.md.
Maybe make this an opt-in by setting `ng-straight-bootstrap-at-startup` to `nil`?

declanqian:
Let's make this optional.

```diff
diff --git a/README.md b/README.md
index d2d7a593fc..646fb86382 100644
--- a/README.md
+++ b/README.md
@@ -125,13 +125,12 @@ We are kind of using `straight.el` as an alternative of `package.el`, providing
 
 ##### Usage
 Emacs NG automatically bootstraps `straight.el` at
-startup if `ng-straight-bootstrap-at-startup` is set to `t`, which is the default.
-You must configure `straight.el` in the early init file,
-as the variable `ng-straight-bootstrap-at-startup` is read before
-loading the regular init file.
+startup if `ng-straight-bootstrap-at-startup` is set to `t`, the default value is `nil`
 
-There are some variables you may be interested in (some of them must be
-set **before** the bootstrap process, if they might affect how
+You must configure the built-in `straight.el` in the early init file,
+as the variable `ng-straight-bootstrap-at-startup` is read before
+loading the regular init file. There are some variables you may be interested in
+(some of them must be set **before** the bootstrap process, if they might affect how
 `straight.el` itself is loaded). You can find the details from [this section](https://github.com/raxod502/straight.el/tree/develop#getting-started)
 of the `straight.el`'s documentation.
 
diff --git a/lisp/emacs-lisp/emacs-ng.el b/lisp/emacs-lisp/emacs-ng.el
index a2e6f0cb33..ed4289c028 100644
--- a/lisp/emacs-lisp/emacs-ng.el
+++ b/lisp/emacs-lisp/emacs-ng.el
@@ -21,7 +21,7 @@ straight-repository-branch
 ;;; Customization options
 
 ;;;###autoload
-(defcustom ng-straight-bootstrap-at-startup t
+(defcustom ng-straight-bootstrap-at-startup nil
   "Whether to bootstrap straight.el when Emacs starts.
 If non-nil, straight.el is bootstrapped before reading the init
 file (but after reading the early init file).  This means that if
```

harryfei:
> > doom-emacs user could choose to disable `ng-straight-bootstrap-at-startup`. Am I right?
> > Can we add this in our doc too?
> 
> There are some occurs in the README.md.
> Maybe make this an opt-in by setting `ng-straight-bootstrap-at-startup` to `nil`?

Frankly, I mean add some specific words for `doom-emacs` issue. If they are facing with this warning, they could find solution in our doc quickly.

declanqian:
How do you configure `early-init` in doom? `ng-straight-bootstrap-at-startup`
has to be set from `early-init.el` stage.
I don't have too much knowledge regarding doom. @harryfei

harryfei:
> How do you configure `early-init` in doom? `ng-straight-bootstrap-at-startup`
> has to be set from `early-init.el` stage.
> I don't have too much knowledge regarding doom. @harryfei

I'm not confirming any solution above in my envrionment. If we can't set `ng-straight-bootstrap-at-startup` early, keeping it optional is good choice.

brotzeit:
> If we can't set ng-straight-bootstrap-at-startup early, keeping it optional is good choice.

I think I agree on this. Maybe we can make this a default later.

PR #386, opened 2021-08-30
webrender: fix multiframe crash issue in wayland platform

harryfei:


rust_src/crates/webrender/src/event_loop.rs
```diff
@@ -1,5 +1,15 @@
 use std::{cell::RefCell, os::unix::prelude::AsRawFd, ptr, sync::Mutex, time::Instant};
 
+#[cfg(macos)]
+use copypasta::osx_clipboard::OSXClipboardContext;
+#[cfg(windows)]
```
brotzeit, line 5: I suppose you just add this for completeness but haven't actually tested it on windows ?

PR #389, opened 2021-08-31
upgrade dependencies.

harryfei:


brotzeit:
I would do the js crate seperately.

harryfei:
I skipped `syn` stuff in this PR @brotzeit too.

declantsien:
I guess this is no longer relevant.

PR #397, opened 2021-09-12
webrender: use thread_select when running without window system

declanqian:
Fixed two minor webrender issues:

- Terminal Emacs (`-nw`) not working on macOS
- Need manually pass `DISPLAY=${hostname}` to open a GUI frame on macOS
Copied the logic from [PGTK](https://github.com/masm11/emacs/blob/pgtk/src/pgtkfns.c#L2127-L2130)

### DIFF
--- a/rust_src/crates/webrender/src/wrterm.rs
+++ b/rust_src/crates/webrender/src/wrterm.rs
@@ -24,10 +24,11 @@ use emacs::{
     bindings::globals,
     bindings::resource_types::{RES_TYPE_NUMBER, RES_TYPE_STRING, RES_TYPE_SYMBOL},
     bindings::{
-        block_input, gui_display_get_arg, hashtest_eql, image as Emacs_Image, list3i, make_fixnum,
-        make_hash_table, make_monitor_attribute_list, register_font_driver, unblock_input, Display,
-        Emacs_Pixmap, Emacs_Rectangle, Fcons, Fcopy_alist, Fmake_vector, Fprovide, MonitorInfo,
-        Vframe_list, Window, DEFAULT_REHASH_SIZE, DEFAULT_REHASH_THRESHOLD,
+        block_input, build_string, gui_display_get_arg, hashtest_eql, image as Emacs_Image, list3i,
+        make_fixnum, make_hash_table, make_monitor_attribute_list, register_font_driver,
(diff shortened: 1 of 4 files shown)


harryfei (approved):

rust_src/crates/webrender/src/wrterm.rs
```diff
@@ -423,6 +424,15 @@ pub fn x_open_connection(
     _xrm_string: LispObject,
     _must_succeed: LispObject,
 ) -> LispObject {
+    let display = if unsafe { NILP(display) } {
+        let display = CString::new("").unwrap();
+        unsafe { build_string(display.as_ptr()) }
```
brotzeit, line 429: `build_string("".as_ptr() as *const ::libc::c_char)` ?

rust_src/crates/webrender/src/wrterm.rs
```diff
@@ -423,6 +424,15 @@ pub fn x_open_connection(
     _xrm_string: LispObject,
     _must_succeed: LispObject,
 ) -> LispObject {
+    let display = if unsafe { NILP(display) } {
```
brotzeit, line 427: `display.is_nil()`

brotzeit: In remacs we've used Option<LispStringRef> as argument in those situations. As far as I remember the function should throw an error if display is not a string or nil.

https://github.com/remacs/remacs/blob/master/rust_src/src/windows.rs#L1196

rust_src/crates/webrender/src/wrterm.rs
declanqian: Thanks

rust_src/crates/webrender/src/wrterm.rs
declanqian: Thanks

rust_src/crates/webrender/src/wrterm.rs
harryfei: Oh, I think this is not safe. IIRC, the `String` is not nul-terminated while `CString` is.

declanqian: Hmm, I thought it's just a code style we're doing. So I just followed your suggestion.
Also I think I was using `CString` in the original code.

brotzeit: Oh, we've used this in remacs several times so I thought it's ok.

PR #400, opened 2021-09-22
also require use-package during bootstrap

brotzeit:
@declanqian it seems it's necessary to also require use-package. I get `nil` when evaluating `(featurep 'use-package)`.

### DIFF
--- a/lisp/emacs-lisp/emacs-ng.el
+++ b/lisp/emacs-lisp/emacs-ng.el
@@ -48,6 +48,7 @@ call (ng-bootstrap-straight) in your init-file."
   (interactive)
 
   (require 'straight)
+  (require 'use-package)
 
   (setq ng--straight-bootstrapped t)
 


black7375 (approved):

declanqian:
Autoloads are always available I think. `(fboundp 'use-package)`.
What is the benefits to requiring it during initialization?

If one user don't use use-package with `straight.el`, this is some extra
`emacs-init-time` I think. @brotzeit 

I would prefer [`striaight.el`'s way](https://github.com/raxod502/straight.el#integration-with-use-package) to integrate `use-package`. We already
[preferred `built-in`](https://github.com/emacs-ng/emacs-ng/blob/master/lisp/emacs-lisp/emacs-ng.el#L89) to reduce some download time.

black7375:
The following code uses `straight.el` to download package `use-package`, so if it's built-in, we can do it like `require` or `fboundp`.
```elisp
(straight-use-package 'use-package)
```

I agree that `fboundp` is better than `require`.

brotzeit:
Sorry for the delay. I think I ran into some issue, but I can't remember =)

It seems to be working now. We probably don't even need `(straight-use-package 'use-package)` ?

declanqian:
I think the catch here is that I removed `(load "emacs-ng")` from `loadup.el` file.
Emacs has some limitation when dumping elisps (lisp preloaded from `loadup.el` file)
So
```
(featurep 'emacs-ng)
(featurep 'use-package)
(featurep 'straight)
``` are all `nil` at the beginning.

declantsien:
This PR is invalid now, since I reverted the use-package/straight.el commits in favor of
use-package from upstream Emacs.
Closing.

PR #401, opened 2021-09-22
introduce wiki section for built-in elisp packages

brotzeit:
@declanqian maybe it's better to keep the readme short and only give an overview of the available features and have more detailed docs in the wiki. What do you think ?


declanqian (approved):
Yeah, it's better organized this way. @brotzeit

PR #409, opened 2021-11-02
ignore errors in nix build

brotzeit:


brotzeit:
@GTrunSec I think it looks better when the tests are green. I would open an issue so we don't forget to fix this.

black7375 (approved):

GTrunSec:
it makes sense.

PR #418, opened 2021-11-17
webrender: integrate wr capture infra

declanqian:
Wrote some rust/lisp, integrating [WebRender capture infrastructure](http://kvark.github.io/webrender/debug/ron/2018/01/23/wr-capture-infra.html) using
Transient interface. Didn't solve any existing problems here.
Only trying to get myself comfortable with the codebase.

Some notable changes:
- Added `webrender-head-rev` lisp var which reflects current webrender version in use.
- Add `message!` macro to send non-error text to Emacs echo area, as a complement to
`error!`.
- Moved webrender auto generated code to webrender/build.rs, aka `colors.rs`. Also newly
added `webrender_revision.rs`.

I think we should avoid generating code for other crates from `rust_src/build.rs`.
Also, did we misused [Cargo workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) for rust_src? Can't get my head around it.
```
rust_src/emacs-ng
rust_src/crates/**
```
Would make more sense to me.

### DIFF
--- a/lisp/wr-fns.el
+++ b/lisp/wr-fns.el
@@ -1,0 +1,176 @@
+;;; wr-fns.el --- Webrender specific functions       -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2021  Declan Qian
+
+;; Author: Declan Qian <>
+;; Keywords: lisp
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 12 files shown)


brotzeit:
> Would make more sense to me.

Agreed. There's also some other stuff we need to update like remacs-lib. I couldn't find an issue, do you want to open one ?

lisp/wr-fns.el
```diff
@@ -0,0 +1,176 @@
...
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;;
+
+;;; Code:
+
+(require 'transient)
```
brotzeit, line 27: Nice, I wasn't aware that transient was merged into emacs :+1:

brotzeit:
LGTM. Let's give @harryfei some time to take a look.

harryfei (approved):
Wow, cool PR.

PR #419, opened 2021-11-18
add macro call that replaces calln functions

brotzeit:


brotzeit:
I have tried it with and without args.

PR #424, opened 2021-12-18
add optional async extension for lsp clients

brotzeit:
This is still very unstable. It seems to depend on the lsp server, the size of a project and the lsp requests. I'll take a deeper look and try to fix it. But I think it makes sense to merge it, so others can also take a look.

The changes depend on https://github.com/emacs-lsp/lsp-mode/pull/2978


appetrosyan:
Is this still necessary given the upstream JSON parser?

PR #425, opened 2021-12-18
use ng_async for magit blame

brotzeit:
I started another try to improve the performance of magit blame. This time I used a subprocess instead of libgit. Instead of doing everything in the main thread this implementation makes use of ng_async for parsing the blame output and additionally get the buffer positions for blame chunks. The latter takes most of the execution time of regular magit blame when using it in large buffers.

However I noticed that emacs can't handle the speed of the incoming chunks when trying to create the overlays for the whole file(at least for big files).

That's why I added a hook for `window-scroll-functions` so every time the window positions change the overlays are created for the visible part of the buffer. This seems to work well, but it's probably necessary to apply some optimizations to make it more stable.

I think ng_async provides lots of possibilities to improve performance even if we can't use `LispObject` in rust threads.


harryfei:
I have give a it a fast and non-serious review. 

I just want to remind you that there still not be dirty commented code.

PR #440, opened 2022-05-31
Merge emacs-28

brotzeit:


black7375 (approved):

brotzeit:
I forgot to copy the diff of the few merge conflicts, but I think it were only 2 spots where upstream has added a word. Should be fine.

PR #461, opened 2023-02-11
Add native fontconfig support

declantsien:
So, before. We use fontdb load_system_fonts to load fonts to use while using fontkit to find
the family name to use.

But fontkit has full fontconfig support while fontdb uses hard-coded path, which would cause problem when fontkit return a name  hasn't been loaded by fontdb.

This patch uses my patched font_loader to get system font dirs for fontdb to use on Unix. Also it
uses font_loaders fontconfig to query font names.

### DIFF
--- a/rust_src/lib.rs
+++ b/rust_src/lib.rs
@@ -1,5 +1,39 @@
+#[cfg(debug_assertions)]
+use tracing_subscriber::{fmt, prelude::*, EnvFilter};
+#[macro_use]
+extern crate emacs;
+
+use emacs::bindings::{main1, terminate_due_to_signal, will_dump_p};
+
 // Include the main c_exports file that holds the main rust_init_syms.
 // This function calls the other crates init_syms functions which contain
 // the generated bindings.
 #[cfg(not(test))]
 include!(concat!(env!("OUT_DIR"), "/c_exports.rs"));
+
+#[no_mangle]
+#[allow(unused_doc_comments)]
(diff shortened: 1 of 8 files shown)


declantsien:
plz review @harryfei

rust_src/lib.rs
```diff
@@ -1,5 +1,39 @@
...
+    // install global collector configured based on RUST_LOG env var.
+    #[cfg(debug_assertions)]
+    tracing_subscriber::registry()
+        .with(fmt::layer())
+        .with(EnvFilter::from_env("EMACSNG_LOG"))
+        .init();
+
+    log::trace!("Emacs NG");
+
+    // tokio::spawn(async move {
```
harryfei, line 32: two unneeded comment line

rust_src/lib.rs
```diff
@@ -1,5 +1,39 @@
+#[cfg(debug_assertions)]
+use tracing_subscriber::{fmt, prelude::*, EnvFilter};
+#[macro_use]
```
harryfei, line 3: rustc says we don't need those two lines. Delete them.

```
#[macro_use]
extern crate emacs;
```

harryfei (changes requested):

harryfei:
This PR works on Arch Linux.

declantsien:
> This PR works on Arch Linux.

I tested on Guix/Nix and macOS. It works for me too.

harryfei (approved):

Issue #442, opened 2022-06-01
Build with js crate using nix failed on darwin


hlolli:
**Describe the bug**

```console
   Compiling swc_ecmascript v0.31.4
   Compiling dprint-swc-ecma-ast-view v0.15.0
   Compiling dprint-swc-ecma-ast-view v0.16.0
   Compiling deno_doc v0.3.0
   Compiling deno_lint v0.4.0
   Compiling deno v1.9.2
   Compiling dprint-plugin-typescript v0.44.2
   Compiling js v0.1.0 (/private/tmp/nix-build-emacsng-1e46579.drv-0/source/rust_src/crates/js)
    Finished dev [unoptimized + debuginfo] target(s) in 3m 16s
sed -i~ 's/rust_src\//..\/rust_src\//g' ../rust_src/target/debug/libemacsng.d
clang -o temacs.tmp \
[...]

Issue #410, opened 2021-11-03
Fix nix build


brotzeit:
We currently ignore errors in the nix build so tests go green.

declantsien:
I've decouple nix flake in a separeted repo here
https://github.com/declantsien/emacs-webrender/actions/workflows/nix-flake.yaml
And it's green on both Linux and macOS.

Issue #312, opened 2021-06-05
macOS nix-build: "make[1]: *** [Makefile:687: ../rust_src/crates/emacs/src/definitions.rs] Killed: 9"


Atemu:
**Describe the bug**

When building emacs-ng on macOS with Web~~reader~~render, something gets killed with signal 9 during the build. I don't think this is an OOM since I've never seen something get reaped on OSX (it does very aggressive swapping).

Getting killed with -9 can sometimes mean that macOS doesn't allow you to be executed because you're lacking an apple-blessed signature or whatever; that might be a lead.

**To Reproduce**
Steps to reproduce the behavior. What were the options you passed to configure?

```patch
diff --git a/flake.nix b/flake.nix
index 073326c19b..579c0082bc 100644
--- a/flake.nix
+++ b/flake.nix
@@ -195,7 +195,7 @@
           librusty_v8 = prev.callPackage ./nix/librusty_v8.nix { };
 
           emacsNg = with prev; let
-            withWebreader = false;
+            withWebreader = true;
           in
           (
             final.emacsGcc.override
@@ -248,6 +248,7 @@
                   "--with-zlib"
                   "--with-dumping=pdumper"
                 ] ++ lib.optionals withWebreader [
+                  "--without-ns"
                   "--with-webrender"
                 ] ++ lib.optionals stdenv.isLinux [
                   "--with-dbus"
```

```shell
git checkout 8c9cf76c03c9aa3f84827b4cc5a4d31a779aebc2
nix-build
```

**Expected behavior**

**OS**
Which os do you use? (`uname -a`)

Darwin MacBook-Pro 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:33 PDT 2021; root:xnu-7195.121.3~9/RELEASE_X86_64 x86_64 i386 MacBookPro15,1 Darwin

**Error callstack**

harryfei:
Webrender feature can't build on macOS currently.

GTrunSec:
Also, I have no idea how to build emacsNg with Nix on darwin. b(Maybe the current problem was from the older version of darwin-SDK)

Atemu:
> Webrender feature can't build on macOS currently.

I see, having the build fail because of a process being killed seemed unusual though.

> I have no idea how to build emacsNg(without webrender) with Nix on darwin. (Maybe the current problem was from the older version of darwin-SDK)

The current non-webrender master gets quite far actually, only fails to link in the end.

[...]

PR #462, opened 2023-02-11
Nix build fix

### LINKED ISSUE #454: Building with Nix Flakes failed
**Describe the bug**
Building with nix flakes failed:
image

**To Reproduce**
I tried to install EmacsNg using nix flakes, but it failed b/c of sha mismatch.

**Expected behavior**
Installing emacsng with nix flakes should work.

**OS**
Nixos

declantsien:
- ~Unclear about the way nix used to work. Nix today requires a hash if we use fetchGit.
Not sure how we pass that hash if the build is triggered by CI.
So instead I just use ./. (the checked out source as src).~

- ~Secondly I see we used to vendor cargo sources using nix on build time. But fetchTarball also
need to specify a hash (fixed-output).~

~There would be a lot of overhead for just updating the hash string.
So instead, I put the vendor-ed source in the repo.
This way, there is just a little overhead for developers when updating dependencies. Instead of
just updating Cargo.toml, they also needs to vendor the sources and update .cargo/config.~

~This is the way how Firefox/chromium are dealing with Rust crates. see:
https://github.com/mozilla/gecko-dev/blob/893a8f062ec6144c84403fbfb0a57234418b89cf/python/mozbuild/mozbuild/vendor/vendor_rust.py
https://github.com/chromium/chromium/tree/main/third_party/rust#rust-third-party-code~

- ~Unclear about the licenses issues though.~

Close #454 #442 #410 #312


GTrunSec:
`vendor cargo sources` is not bad for me.  I also have a refactor work in mind, and maybe someday I will put it on. 

- LGTM

harryfei:
I don't think it's a good idea.

GTrunSec:
We can temporarily turn off the CI of nix-build, and then turn it on when I find free time to refactor it.

declantsien:
> I think it's bad practice. It makes the repo unnecessary huge.

I am hesitated with huge repo size too. I am gonna drop this.

> We can temporarily turn off the CI of nix-build,

Disabled using github interface. See here
https://github.com/emacs-ng/emacs-ng/actions/workflows/checks.yaml

> and then turn it on when I find free time to refactor it.

Looking forward to it.

declantsien:
Today, I realized that nix has support for automatically [vendoring source using Cargo.lock file](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md#vendoring-of-dependencies-vendoring-of-dependencies).
I tried it out. It works!
Also I removed the formatting changes. Would you like to give this a review? @GTrunSec

GTrunSec:
> https://github.com/declantsien/emacs-ng/blob/nix-build-fix/flake.nix#L152
nice, we did the same thing.

GTrunSec (approved):

Atemu:
This is great, thanks!

PR #479, opened 2023-02-22
feat(WR): add alternative pselect support

declantsien:
~This PR dropped tokio dependency for Emacs WR by leveraging pselect over tokio async Fd.~
This PR adds select as an alternative to tokio AsyncFD for WR synchronous I/O multiplexing.
pselect is used on darwin platform. On other platform, tokio is used by default. However, we
can also enable plain select with `--enable-wr-pselect` if users feel like to.

The motivation for me to make this change is that:
When using Rust future (async Fd?), which caused a lot of crash on macOS
last time I tested.

With pselect, I haven't noticed performance difference.


harryfei:
Could you just keep it on Linux platform and use plain select on macOS. tokio give us ability to make emacs run in multiple threads based concurrency way, though we haven't use it just now.

declantsien:
Harry ***@***.***> writes:

> Could you just keep it on Linux platform and use plain select on macOS. tokio give us ability to make emacs run in multiple threads based concurrency way, though we haven't use it just now. 
>
> -- 
> https://github.com/emacs-ng/emacs-ng/pull/479#issuecomment-1439878964
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>

Ok. I'll do that.

You don't have crashes runing with current code? I remember I have similar crashes on Linux
`do_one_unbind` from `eval.c` trying to access a 0x0 pointer.

harryfei:
No. Maybe some your config trigger the unimplemented part of emacs-ng webrender while mine don't.
Besides, I'm not using emacs-ng master(merge emacs 29). The new emacs 29 may also break our webrender backend which occurs in past upstream merge.

declantsien:
Harry ***@***.***> writes:

> No. Maybe some your config trigger the unimplemented part of emacs-ng webrender while mine don't.
> Besides, I'm not using emacs-ng master(merge emacs 29). The new emacs 29 may also break our webrender backend which occurs in past upstream merge.
>
> -- 
> https://github.com/emacs-ng/emacs-ng/pull/479#issuecomment-1439976536
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>

Have Not experienced with new issues with the newly upstream master
merge other than unimplemented issues that are well known already.

Are you not using native comp or the newly introduced treesit from
upstream?

harryfei:
Oh, maybe I'm using the `-with-native-compilation` argument which should be `--with-native-compilation`. 
I'm not sure whether I'm using a native comp version now.

declantsien:
> [I had turned it off]

Having problems running with native comp enabled? You can check availability with
`(native-comp-available-p)`
`(treesit-available-p)`

declantsien:
> Could you just keep it on Linux platform and use plain select on macOS. tokio give us ability to make emacs run in multiple threads based concurrency way, though we haven't use it just now.

Updated. Can you give it a look? @harryfei

2e0byo:
> > [I had turned it off]
> 
> Having problems running with native comp enabled? You can check availability with `(native-comp-available-p)` `(treesit-available-p)`

I think `(native-comp-available-p)` still returns true if `native-comp-deferred-compilation` is nil.  I had forgotten I'd set that.  With nativecomp enabled it's even faster! :D

harryfei (commented):
LGTM

2e0byo:
@harryfei  can you reproduce with help as per #506 ?

PR #496, opened 2023-03-01
feat(WR): use webrender with PGTK

declantsien:
This PR

- Added gtk3 gl_area support for winit (tao) build.(`--with-winit=t --with-wr-gl=gt`)
- Added initial support for build PGTK with Webrender. (`--with-pgtk --with-webrender`)

For the PGTK with Webrender. It kind of just works.
I've tested the child-frame/context-menu. Since child-frame is the one I want most when using WR.


p00f:
childframes are white for me:
image
they are white even in dark theme:
image


I used `--with-pgtk --with-webrender`

PR #501, opened 2023-03-02
Fix winit scaling issues

### LINKED ISSUE #493: Winit/WR scaling code is broken
Issue created based on [comment](https://github.com/emacs-ng/emacs-ng/issues/488#issuecomment-1445407191)

> the scaling code is broken, but it always was with winit. (Causes some weird bug with the cursor as clearly two different values for scale are being used internally.)

It should be specific when using winit as windowing implementation.

declantsien:
Close #493

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1628,7 +1628,7 @@ dependencies = [
 [[package]]
 name = "glsl-to-cxx"
 version = "0.1.0"
-source = "git+https://github.com/servo/webrender.git?rev=a578ce7ebf7ae3f50948d1620c7a41d422cc3bfd#a578ce7ebf7ae3f50948d1620c7a41d422cc3bfd"
+source = "git+https://github.com/servo/webrender.git?rev=23f5f305624bd00f450407180fdef74796bd3964#23f5f305624bd00f450407180fdef74796bd3964"
 dependencies = [
  "glsl",
 ]
@@ -2828,7 +2828,7 @@ checksum = "d01a5bd0424d00070b0098dd17ebca6f961a959dead1dbcbbbc1d1cd8d3deeba"
 [[package]]
 name = "peek-poke"
 version = "0.3.0"
-source = "git+https://github.com/servo/webrender.git?rev=a578ce7ebf7ae3f50948d1620c7a41d422cc3bfd#a578ce7ebf7ae3f50948d1620c7a41d422cc3bfd"
+source = "git+https://github.com/servo/webrender.git?rev=23f5f305624bd00f450407180fdef74796bd3964#23f5f305624bd00f450407180fdef74796bd3964"
 dependencies = [
  "euclid",
  "peek-poke-derive",
@@ -2837,7 +2837,7 @@ dependencies = [
 [[package]]
 name = "peek-poke-derive"
 version = "0.3.0"
-source = "git+https://github.com/servo/webrender.git?rev=a578ce7ebf7ae3f50948d1620c7a41d422cc3bfd#a578ce7ebf7ae3f50948d1620c7a41d422cc3bfd"
+source = "git+https://github.com/servo/webrender.git?rev=23f5f305624bd00f450407180fdef74796bd3964#23f5f305624bd00f450407180fdef74796bd3964"
 dependencies = [
  "proc-macro2 1.0.49",
  "quote 1.0.23",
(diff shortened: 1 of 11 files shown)


2e0byo:
This breaks mouse position when scale != 1.

declantsien:
> This breaks mouse position when scale != 1.

Ive tested it with
`
swaymsg output eDP-1 scale 1`
swaymsg output eDP-1 scale 2`
`
Both are working fine. Are you stil having issue with winit/tao? @2e0byo

2e0byo:
Yes.  Mouse position with tao is off by the scale factor.  Funny as this looks like doing the scaling properly.  Under Hyprland; I'll try with weston in a moment.

declantsien:
2e0byo ***@***.***> writes:

> Yes.  Mouse position with tao is off by the scale factor.  Funny as this looks like doing the scaling properly.  Under Hyprland; I'll try with weston in a moment.
>
> -- 
> https://github.com/emacs-ng/emacs-ng/pull/501#issuecomment-1451713680
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>

Ok, Just tried again with tao. It does have issue. Will fix when I got time.

declantsien:
2e0byo ***@***.***> writes:

> Yes.  Mouse position with tao is off by the scale factor.  Funny as this looks like doing the scaling properly.  Under Hyprland; I'll try with weston in a moment.
>
> -- 
> https://github.com/emacs-ng/emacs-ng/pull/501#issuecomment-1451713680
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>

This[0], the latest should fix tao. 


[0] https://github.com/emacs-ng/emacs-ng/compare/1055b432c4d2ae865b88279c3417919dbb330c2e..a092673b5ff434f605b08ca73692bf9aa3a404a3

2e0byo:
> This[0], the latest should fix tao. [0] https://github.com/emacs-ng/emacs-ng/compare/1055b432c4d2ae865b88279c3417919dbb330c2e..a092673b5ff434f605b08ca73692bf9aa3a404a3

It does, but where on earth is `OutputInner.cursor_position` actually defined?  I feel thoroughly stupid.  Anyhow it works.

declantsien:
2e0byo ***@***.***> writes:

>> This[0], the latest should fix tao. [0] https://github.com/emacs-ng/emacs-ng/compare/1055b432c4d2ae865b88279c3417919dbb330c2e..a092673b5ff434f605b08ca73692bf9aa3a404a3
>
> It does, but where on earth is `OutputInner.cursor_position` actually defined?  I feel thoroughly stupid.  Anyhow it works.
>
>
>
> -- 
> https://github.com/emacs-ng/emacs-ng/pull/501#issuecomment-1451802810
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>

It's here `rust_src/crates/webrender/src/window_system/winit/output.rs`.
Nah, I am feeling that our code may be somehow over-engineered.

2e0byo:
In this case I really was being daft, I'd missed that it's just a struct member rather than a function... and then it's set from the cursor move event in `winit/term.rs`. Which is much better than polling anyhow.  thank you!

2e0byo:
We still need to handle [`ScaleFactorChanged`](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.ScaleFactorChanged) events (e.g.if a window is dragged from one monitor to another).   The event gets the new size so it might be as simple as calling `output.resize()`, or we might need to trigger a rebuild.

declantsien:
2e0byo ***@***.***> writes:

> In this case I really was being daft, I'd missed that it's just a struct member rather than a function... and then it's set from the cursor move event in `winit/term.rs`. Which is much better than polling anyhow.  thank you!
>
> -- 
> https://github.com/emacs-ng/emacs-ng/pull/501#issuecomment-1451847181
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>

This is probably a bug from tao. They have `cursor_position` associated
with `Window` which has type `PhysicalPosition`. I expected they update
the value as the one we now received from `CursorMove` event. Apparently
it not the same when `scale_factor` been other than 1. Anyway, I don't
want to dig to the bottom of it. Glad it's working now.

2e0byo:
The code paths are actually completely independent.  I opened an issue at tao: https://github.com/tauri-apps/tao/issues/708

PR #503, opened 2023-03-02
fix: avoid window system when running in terminal

### LINKED ISSUE #502: `emacs -nw` calls window code
**Describe the bug**
`emacs -nw` calls `wr_adjust_canvas_size()` despite webrender not running.  This results (under tao on my hardware) in a segfault when tao tries to get a non-existent window.  This also causes `emacs -nw` to fail when a graphical environment can't be found, e.g. over ssh.

**To Reproduce**

```shell
./configure --with-winit=tao --with-wr-gl=glutin
```

**Expected behavior**
Emacs should not call any graphical code when launching in a terminal.

**OS**
Which os do you use? (`uname -a`)

`Linux 6.2.1-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 26 Feb 2023 03:39:23 +0000 x86_64 GNU/Linux`

**Error callstack**

```shell
Program received signal SIGSEGV, Segmentation fault.
                                                    0x0000555555ab7bdc in wrterm::display_info::DisplayInfo::get_inner::haca182995104828c ()
(gdb) bt
#0  0x0000555555ab7bdc in wrterm::display_info::DisplayInfo::get_inner::haca182995104828c ()
#1  0x0000555555ac4bce in wrterm::output::Canvas::build::h57abddb1fff9f40a ()
#2  0x0000555555ac4ac8 in <emacs::lisp::ExternalPtr<emacs::bindings::frame> as wrterm::frame::LispFrameExt>::canvas::heb8da1159cdfb60c ()
#3  0x0000555555b0fcfe in wr_adjust_canvas_size ()
#4  0x0000555555600fe0 in adjust_frame_size
    (f=0x5555563b6370, new_text_width=<optimized out>, new_text_height=<optimized out>, inhibit=inhibit@entry=5, pretend=<optimized out>, parameter=parameter@entry=0x4110) at frame.c:910
#5  0x00005555555f4ad5 in change_frame_size_1
    (f=<optimized out>, new_width=<optimized out>, new_height=<optimized out>, pretend=pretend@entry=false, delay=delay@entry=false, safe=safe@entry=true)
    at dispnew.c:5984
#6  0x00005555555fd6e3 in change_frame_size (safe=true, delay=false, pretend=false, new_height=<optimized out>, new_width=<optimized out>, f=<optimized out>)
    at dispnew.c:6017
#7  init_display_interactive () at dispnew.c:6621
#8  init_display () at dispnew.c:6672
#9  0x00005555556d8996 in main1 (argc=<optimized out>, argv=0x7fffffffe4a8) at emacs.c:2473
#10 0x0000555555814211 in main ()
(gdb)
```

2e0byo:
closes #502 .

I've also dropped the explicit test for `HAVE_PGTK` here, as it doesn't make sense to call `wrterm` functions unless we have webrender.  I don't know if you can still build non-webrender pgtk with this fork, but it seems worth leaving the possibility in place if we can.

*WHY* isn't it documented anywhere that frame->parent_frame being non-nil is the test for a graphical frame?!  Took me a good twenty minutes to persuade myself it really was always true.

More interestingly, why did this work before?  `emacs -nw` always failed if it couldn't get the graphical display, but other than that it was fine.  This way it works over ssh again :D

### DIFF
--- a/rust_src/crates/webrender/src/term.rs
+++ b/rust_src/crates/webrender/src/term.rs
@@ -24,12 +24,11 @@ use emacs::{
     },
     bindings::{
         draw_fringe_bitmap_params, fontset_from_font, glyph_row, glyph_string, terminal,
-        text_cursor_kinds, Emacs_Color, Emacs_Pixmap, Fprovide, Fredisplay, Fredraw_display,
-        Fredraw_frame,
+        text_cursor_kinds, Emacs_Color, Emacs_Pixmap, Fprovide,
     },
     font::LispFontRef,
     frame::{LispFrameRef, Lisp_Frame},
-    globals::{Qnil, Qt, Qwr},
+    globals::{Qnil, Qwr},
     glyph::GlyphStringRef,
     lisp::{ExternalPtr, LispObject},
     window::{LispWindowRef, Lisp_Window},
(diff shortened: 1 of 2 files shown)


declantsien:
That would be a mistake I made. It should be
`#if defined (USE_WEBRENDER) && defined (HAVE_PGTK)`
Also there is `inhibit_window_system` for checking whether window system is involved.
Can you try with that?

declantsien:
> WHY isn't it documented anywhere that frame->parent_frame being non-nil is the test for a graphical frame?

I doubt that this can be use to check for graphical frame. You can always check the [source code](https://github.com/emacs-ng/emacs-ng/blob/master/src/frame.h#L100).

I think we should check [`output_method`](https://github.com/emacs-ng/emacs-ng/blob/master/src/frame.h#L381) here.

Added some changes to your code:
- Removed line `if (!NILP (f->parent_frame))`, and added `&& HAVE_PGTK`.
- More checking from Rust cide

2e0byo:
Pretty sure it acts as an indirect test.  Parent frame is unset even when using `C-x 5 2` to make a new frame on a terminal.  But rust side checking is more explicit and definitely works, and anyhow this behaviour wasn't documented anywhere.  But it is the test which frame.c relies on a bit earlier when resizing x frames.

declantsien:
2e0byo ***@***.***> writes:

> Pretty sure it acts as an indirect test.  Parent frame is unset even when using `C-x 5 2` to make a new frame on a terminal.  But rust side checking is more explicit and definitely works, and anyhow this behaviour wasn't documented anywhere.  But it is the test which frame.c relies on a bit earlier when resizing x frames.
>
> -- 
> https://github.com/emacs-ng/emacs-ng/pull/503#issuecomment-1453235491
> You are receiving this because you modified the open/close state.
>
> Message ID: ***@***.***>

Good to know. Since only canvas used with PGTK needs to explicitly
resized. I limited that code only run in WR/PGTK. Also made sure its
working for `emacs --daemon` usage.

Sorry to modify your code like that without consultation.

2e0byo:
No problems, glad it's fixed.

I prefer the rust test anyway as I dream (foolishly) of gradually lifting *all* the display code out of c and documenting it properly in the process.

PR #508, opened 2023-03-04
fix(WR): fix WR/PGTK scaling issue

### LINKED ISSUE #500: WR/PGTK handle dpi/scale factor

declantsien:
Close #500


declantsien:
2e0byo ***@***.***> writes:

> Works for me (TM).  I see you scaled up the emacs frame in the end.  I wonder if that's related to emacs consuming 100% cpu if the window is too large? But I need to profile there.
>

Still working on it. Maybe it the font scaling code. We call gtk foreign
function to get the scale factor too often.

> Occasionally dies with failed memory allocation (~8GB are available)
> but frustratingly hard to track down.

I have that too.

2e0byo:
>  We call gtk foreign function to get the scale factor too often.
I don't *think* this is it---I had the same problem before when I hard coded it.

>> Occasionally dies with failed memory allocation (~8GB are available) but frustratingly hard to track down.
> I have that too.

Git bisect blames 3b3f901983f0bf272eb620ea0e7b5bce0e499ddc but it's frustratingly difficult to bisect properly.

Separately I think 59eadeaa08d00c313eaa016989fa4e00bb29ec11 needs reverting, but again I need to confirm properly.

Issue #498, opened 2023-03-01
Pure Rust font stack for wr_glyph_rasterizer


declantsien:
https://github.com/emacs-ng/emacs-ng/issues/498#issuecomment-1490348713

2e0byo:
Isn't this down to automatic composition [not being implemented](https://github.com/emacs-ng/emacs-ng/blob/master/rust_src/crates/webrender/src/draw_canvas.rs#L434) ?

declantsien:
2e0byo ***@***.***> writes:

> Isn't this down to automatic composition [not being implemented](https://github.com/emacs-ng/emacs-ng/blob/master/rust_src/crates/webrender/src/draw_canvas.rs#L434) ?
>
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/emacs-ng/emacs-ng/issues/498#issuecomment-1450601771
> You are receiving this because you authored the thread.
>
> Message ID: ***@***.***>

Seems like that is the case.

declantsien:
Hit this WR [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1565588). WebRender currently does not work with `Noto Color Emoji`.
```
2023-03-15T02:04:03.872358Z ERROR wr_glyph_rasterizer::platform::unix::font: Unable to set glyph size and transform: 23
2023-03-15T02:04:03.872373Z DEBUG wr_glyph_rasterizer::platform::unix::font:    [662] for size 16.0 and scale (1.0, 1.0) from font FontKey(IdNamespace(1), 5)
2023-03-15T02:04:03.872393Z ERROR wr_glyph_rasterizer::platform::unix::font: Unable to set glyph size and transform: 23
2023-03-15T02:04:03.872404Z DEBUG wr_glyph_rasterizer::platform::unix::font:    [260158] for size 16.0 and scale (1.0, 1.0) from font FontKey(IdNamespace(1), 5)```

declantsien:
Code path for color font is not limited to glyph automatic composite.
It used for glyph_type::CHAR_GLYPH too.

Better if WebRender [GlyphInstance](https://doc.servo.org/webrender_api/struct.GlyphInstance.html) adds supports for this, otherwise we have to manually draw font png/svg images.

While waiting for WebRender [adding support for Color Font](https://bugzilla.mozilla.org/show_bug.cgi?id=1565588). I am happy with just disable Noto Color Emoji for now.

declantsien:
[...]

PR #522, opened 2023-03-18
Wr disable noto color emoji

declantsien:
WebRender's [wr_glyph_rasterizer](https://github.com/servo/webrender/tree/master/wr_glyph_rasterizer) currently is unable to handle color fonts like "Noto Color Emoji". See [#498](https://github.com/emacs-ng/emacs-ng/issues/498). We disable "Noto Color Emoji" for WR renderer in this PR.

Also included some code refactor:

- Added debug trait for `LispObject`.
So we can debug `LispObject` obj with `println!({obj:?})`
- Added `From/To<String>` trait for `LispObject`.
- Implemented overstrike face for WR renderer.
- rename `ttf_parser` to `tffp` according to their [c-api](https://github.com/RazrFalcon/ttf-parser/blob/master/c-api/ttfparser.h)

### DIFF
--- a/rust_src/crates/emacs/src/lisp.rs
+++ b/rust_src/crates/emacs/src/lisp.rs
@@ -2,17 +2,19 @@
 //! lisp.h.
 
 use std::ffi::CString;
+use std::fmt;
 use std::mem;
 use std::ops::{Deref, DerefMut};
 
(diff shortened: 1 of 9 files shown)


2e0byo:
hurrah for the debug trait, better than my hacks locally :D

PR #524, opened 2023-03-24
Add more image formats support

declantsien:
- Add svg and all formats that image_rs supports by default.
- Add glyph automatic composition

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2059,6 +2059,14 @@ fi
 
 if test "${with_webrender}" = "yes" ; then
   with_cairo=no
+dnl WebRender support these images from native image support
+  with_xpm=no
+  with_jpeg=no
+  with_tiff=no
+  with_gif=no
+  with_png=no
+  with_rsvg=no
+  with_webp=no
 fi
 
 LD_SWITCH_X_SITE_RPATH=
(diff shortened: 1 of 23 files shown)


configure.ac
2e0byo, line 1: Nit: current implementation doesn't support xpm as there's no rust xpm decoder.  I do actually have a hack to support it via imagemagick I've been meaning to PR, but it would be better just to write a decoder crate for image.rs (it's a trivial format and the spec is only a few pages).

Maybe worth adding a separate comment for `xpm`.  but we can't use emac's `libxpm` anyway without ffi, so I agree there's no point building with xpm support.

harryfei:
Wow. This is what I expect to have when I choose `image-rs` in the initial webrender image implementation.

emacs-ng is more powerful than vinilla emacs now. It even supports `avif` !

declantsien:
Harry ***@***.***> writes:

> Wow. This is what I expect to have when I choose `image-rs` in the initial webrender image implementation.
>
> emacs-ng is more powerful than vinilla emacs now. It even supports `avif` !

Oh, avif is not in image-rs's Cargo default feature and requires native
dependency `libdav1d`. So I left it out for now.

FYI, I don't know too know about images formats(standards).

Issue #336, opened 2021-06-24
Build fails on apple silicon (M1/aarch64-darwin) 


shaunsingh:
Hi, I was trying to build the application on my m1 machine, and got the following error/output 

Both autogen and configure work, it fails on the `make -j 8` step 

```zsh
emacs-ng on  master [!] took 37s
❯ make -j 8
/Library/Developer/CommandLineTools/usr/bin/make -C lib all
/Library/Developer/CommandLineTools/usr/bin/make -C doc/lispref info
/Library/Developer/CommandLineTools/usr/bin/make -C doc/lispintro info
/Library/Developer/CommandLineTools/usr/bin/make -C doc/emacs info
../.././build-aux/install-sh -c -d ../../info
../.././build-aux/install-sh -c -d ../../info
../.././build-aux/install-sh -c -d ../../info
  GEN      alloca.h
  GEN      byteswap.h
  GEN      dirent.h
  GEN      fcntl.h
  GEN      getopt.h
  GEN      getopt-cdefs.h
  GEN      ieee754.h
  GEN      inttypes.h
echo '#include "mini-gmp.h"' >gmp.h-t
  GEN      limits.h
mv gmp.h-t gmp.h
  GEN      signal.h
  GEN      stdint.h
  GEN      stdio.h
  GEN      stdlib.h
  GEN      string.h
mkdir: ../../info: File exists
mkdir: ../../info: File exists
mkdir: ./info: File exists
  GEN      sys/random.h
  GEN      ../../info/emacs.info
  GEN      ../../info/eintr.info
  GEN      ../../info/elisp.info
  GEN      info/dir
  GEN      sys/select.h
  GEN      sys/stat.h
  GEN      sys/types.h
  GEN      sys/time.h
  GEN      time.h
  GEN      unistd.h
  CC       fingerprint.o
  CC       acl_entries.o
  CC       canonicalize-lgpl.o
  CC       copy-file-range.o
  CC       euidaccess.o
  CC       explicit_bzero.o
  CC       faccessat.o
  CC       fpending.o
  CC       free.o
  CC       fstatat.o
  CC       fsusage.o
  CC       futimens.o
  CC       getgroups.o
  CC       getopt.o
  CC       getopt1.o
  CC       getrandom.o
  CC       group-member.o
  CC       lstat.o
  CC       memmem.o
  CC       mempcpy.o
  CC       memrchr.o
  CC       mini-gmp-gnulib.o
  CC       mktime.o
[...]

PR #541, opened 2024-02-11
Add aarch64-darwin to flake

sempruijs:
I wanted to check if this product is something for me, so I tried to run:

```
nix run github:emacs-ng/emacs-ng
```

I forked the repo to check what error the build would give me if I try to compile this program on my machine.
There was no error and everything compiled just fine. I noticed #336 but that error showed up with the makefile if I'm not mistaken.
Nix is reproducible so you can test it yourself ;)

It's my first time here, feedback is appreciated :))

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -30,7 +30,7 @@
   }:
     {}
     // (
-      flake-utils.lib.eachSystem ["x86_64-linux" "x86_64-darwin"]
+      flake-utils.lib.eachSystem ["x86_64-linux" "x86_64-darwin" "aarch64-darwin"]
       (
         system: let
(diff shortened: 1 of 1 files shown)


declantsien:
Are you using macOS M1? I assumed `aarch64-darwin` is for ARM macs.
We currently do have Github action for macOS M1. I think it's OK to add
another output for Nix flake though.

sempruijs:
Yes I have a ARM mac. It is usefull for me because I manage my mac with nix darwin. If this platform is added I can use this flake as an input so I can have the most up-to-date emacs-ng

PR #568, opened 2024-03-07
House keeping

declantsien:
* Move window system traits to emacs crates
* Cleanup unused code for font crate
* Remove tao related code
Gtk is slow, I can feel it. So removing tao which is implemented using gtk on linux.
Leaving Pgtk with wr as a reference.


declantsien:
> Out of interest, what do you use? I gave up on emacs-ng as a daily driver when the thesis hit; I'd like to go back to it now.

The latest commit on master branch causing high cpu usage. I reverted it for now.
Other than that, I found winit term is stable enough. I am using winit term on Wayland
only recently.

I am planning add webview(wry with servo) and [wayland child](https://github.com/rust-windowing/winit/pull/3487) frame support though.

Issue #20, opened 2020-11-19
Move the generated bindings in separate project


yyoncho:
The generated bindings are 6mb file and they pretty much kill all the IDE tools and `cargo check` (e. g. each change triggers recompilation of the bindings).  This PR is POC https://github.com/emacs-ng/emacs-ng/pull/12 and proves that it can be done. Ideally, no other code except the bindings should be in that project.

brotzeit:
The problem here is that we end up with cyclic dependencies(with the lisp crate) when we move the generated bindings in a seperate crate while using `LispObject` in the bindings crate. 

My first thought is that we use `Lisp_Object` instead. But we then would have to perform `obj.into()` in each call of generated functions. Do we want that ? Are there any other possible solutions ?

appetrosyan:
```
obj.into()
```
is a small price to pay for the improved compilation times. And we can make use of `as_ref` in most cases to elide the need for conversions. From the point of view of decoupling it might even be preferable to avoid a circular dependency by opacifying the object.

declantsien:
I guess the name of this separate project would be `emacs-sys`,
like the `gtk-sys`, 'llvm-sys', which provides bindings to Emacs's C API.
And some Rust traits for those function/macros defined in C header files.

The generated bindings is in Cargo `OUT_DIR`, since rust-analyzer now
supports jumping to file in `OUT_DIR`.

We put this crate inside the source tree instead of a separate project.

PR #599, opened 2024-03-25
Reorganize Rust source

declantsien:
* move all crates to root/crates dirs like other Rust projects
* Rename emacs crate to emacs-sys like other *-sys crates in cargo registry
* Rename Remacs-lib to lisp-doc, as it's only use currently is to
support lisp documentation functionality for those lisp fn/var/symbols written
in rust
* Rename ng-bindgen to codegen, as the bindgen is called from emacs-sys
build script. And the purpose of this crate is to generate code like
`c_exports.rs` and other


declantsien:
Close #23 , Close #20

PR #639, opened 2025-01-26
Pull upstream 35d3927

declantsien:


declantsien:
Nix CI on Linux/macOS is actually working except the nix cache step which is from
runner itself.

Normal CI(Linux and macOS) has this error
"  package `home v0.5.11` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.77.0-nightly"
Urging us to update rust-toolchain which is annoying.

Nix wins here. I'll update rust-toolchain later.

